PR #1, opened 2016-11-03
Export bsearch

klaussfreire:


sharedbuffers/mapped_struct.py
```diff
@@ -1760,6 +1823,252 @@ def _c_search_hkey_ui32(hkey, pindex, stride0, length, hint):
...
+        lo = 0
+        if lo < hi:
+            # First iteration a quick guess assuming uniform distribution of keys
+            mid = min(hint, hi-1)
+            mkey = cython.cast(cython.p_double, pindex + stride0 * mid)[0]
+            if mkey < hkey:
+                # Got a lo guess, now skip-search forward for a hi
+                lo = mid = mid+1
+                skip = 32
+                while skip > 0 and mid + skip < hi:
```
zeitos, line 1899: I don't think you need to check skip. Are you worried about overflow?

sharedbuffers/mapped_struct.py
```diff
@@ -1708,6 +1712,65 @@ def _c_search_hkey_ui64(hkey, pindex, stride0, length, hint):
...
+                        skip *= 2
+                    else:
+                        hi = mid + skip
+                        break
+            elif mkey > hkey:
+                # Got a hi guess, now skip-search backwards for a lo
+                hi = mid
+                skip = 32
+                while skip > 0 and mid > lo + skip:
+                    if cython.cast(cython.p_longlong, pindex + stride0 * (mid-skip))[0] > hkey:
```
zeitos, line 1745: Same

sharedbuffers/mapped_struct.py
```diff
@@ -1708,6 +1712,65 @@ def _c_search_hkey_ui64(hkey, pindex, stride0, length, hint):
...
+                    mid -= 1
+                return mid
+        # Final stretch: search the remaining range with a regular binary search
+        while lo < hi:
+            mid = (lo+hi)//2
+            mkey = cython.cast(cython.p_longlong, pindex + stride0 * mid)[0]
+            if mkey < hkey:
+                lo = mid+1
+            elif mkey > hkey:
+                hi = mid
```
zeitos, line 1765: You are not using hi

tests/mapped_struct.py
```diff
@@ -591,3 +593,28 @@ class MappedStringMappingBigTest(MappedStringMappingTest):
 class MappedString32MappingBigTest(MappedString32MappingTest):
     TEST_KEYS = [ "%d%s" % (i,k) for k in MappedStringMappingTest.TEST_KEYS for i in xrange(64) ]
     TEST_VALUES = [ v for v in MappedStringMappingTest.TEST_VALUES for i in xrange(64) ]
+
+class BsearchTest(unittest.TestCase):
+    SUPPORTED_DTYPES = [ numpy.uint32, numpy.int32, numpy.uint64, numpy.int64,
```
tzulberti-jampp, line 598: It would be great to have a not supported type to check what it might happen in that case. Make sure that the exception in raised

sharedbuffers/mapped_struct.py
```diff
@@ -1760,6 +1823,252 @@ def _c_search_hkey_ui32(hkey, pindex, stride0, length, hint):
...
+        hkey = cython.float,
+        mkey = cython.float,
+        lo = cython.size_t, hi = cython.size_t, length = cython.size_t,
+        mid = cython.size_t, mid2 = cython.size_t, stride0 = cython.size_t, hint = cython.size_t,
+        pindex = cython.p_char, skip = cython.size_t)
+    @cython.returns(cython.size_t)
+    def _c_search_hkey_f32(hkey, pindex, stride0, length, hint):
+        hi = length
+        lo = 0
+        if lo < hi:
```
tzulberti-jampp, line 1950: When could it happen that `lo >= hi`? I understand that you are making sure that the array isn't empty `0 < len(a)`, but `a` can not be empty because the caller, `hinted_bsearch`, already uses the position 0 of `a` so it can not be empty.

sharedbuffers/mapped_struct.py
```diff
@@ -1875,6 +1877,124 @@ def _c_search_hkey_i32(hkey, pindex, stride0, length, hint):
...
 
+    @cython.cfunc
+    @cython.locals(
+        hkey = cython.double,
+        mkey = cython.double,
+        lo = cython.size_t, hi = cython.size_t, length = cython.size_t,
+        mid = cython.size_t, mid2 = cython.size_t, stride0 = cython.size_t, hint = cython.size_t,
+        pindex = cython.p_char, skip = cython.size_t)
+    @cython.returns(cython.size_t)
+    def _c_search_hkey_f64(hkey, pindex, stride0, length, hint):
```
zeitos, line 1888: Can you document what's the intention of this method?

zeitos: Also: can you add a test?

zeitos (changes requested):

sharedbuffers/mapped_struct.py
```diff
@@ -1708,6 +1712,65 @@ def _c_search_hkey_ui64(hkey, pindex, stride0, length, hint):
...
+        lo = cython.size_t, hi = cython.size_t, length = cython.size_t,
+        mid = cython.size_t, mid2 = cython.size_t, stride0 = cython.size_t, hint = cython.size_t,
+        pindex = cython.p_char, skip = cython.size_t)
+    @cython.returns(cython.size_t)
+    def _c_search_hkey_i64(hkey, pindex, stride0, length, hint):
+        hi = length
+        lo = 0
+        if lo < hi:
+            # First iteration a quick guess assuming uniform distribution of keys
+            mid = min(hint, hi-1)
```
zeitos, line 1726: I don't think you need this min

sharedbuffers/mapped_struct.py
klaussfreire: The caller isn't only hinted_bsearch (well, right now it is, but NumericIdMapper could in the future call it).

klaussfreire: Besides, those checks are usually also a good idea since they give information to the compiler (ie: now the compiler knows lo < hi and can - and will - optimize accordingly)

sharedbuffers/mapped_struct.py
klaussfreire: The whole test suite tests these methods. There's an explicit test in testBsearchFloat64.

It's a binary search for doubles.

sharedbuffers/mapped_struct.py
klaussfreire: Yes I am. `while lo < hi`

sharedbuffers/mapped_struct.py
klaussfreire: Yep. Overflow can and does happen. Making a test for it requires a 32-bit system though.

sharedbuffers/mapped_struct.py
klaussfreire: hint is an externally supplied index. I do need it to avoid a segmentation fault.

sharedbuffers/mapped_struct.py
klaussfreire: BTW: the first copy of those (there's tons of copies, I didn't find a way to avoid it) has comments about why it does things the way it does. It's not on the diff since I didn't touch it, I just added more copies for more data types

klaussfreire: Ping?

tests/mapped_struct.py
```diff
@@ -591,3 +593,40 @@ class MappedStringMappingBigTest(MappedStringMappingTest):
...
+            a = numpy.sort(a)
+            for x in a:
+                ix = mapped_struct.bsearch(a, x)
+                self.assertLess(ix, len(a))
+                self.assertEqual(a[ix], x)
+                self.assertTrue(mapped_struct.sorted_contains(a, x))
+            for x in b:
+                ix = mapped_struct.bsearch(a, x)
+                self.assertTrue(ix >= len(a) or a[ix] != x)
+                self.assertTrue(not mapped_struct.sorted_contains(a, x))
```
tzulberti-jampp: In this case, isn't this the same to `assertFalse(mapped_struct.sorted_contains(a, x))`?

tests/mapped_struct.py
```diff
@@ -591,3 +593,40 @@ class MappedStringMappingBigTest(MappedStringMappingTest):
...
+            b = numpy.array(testarray[50:], dtype)
+            a = numpy.sort(a)
+            for x in a:
+                ix = mapped_struct.bsearch(a, x)
+                self.assertLess(ix, len(a))
+                self.assertEqual(a[ix], x)
+                self.assertTrue(mapped_struct.sorted_contains(a, x))
+            for x in b:
+                ix = mapped_struct.bsearch(a, x)
+                self.assertTrue(ix >= len(a) or a[ix] != x)
```
tzulberti-jampp, line 620: I think that special cases like:
- `x > max(a) <=> ix == len(a)`
- `x < min(b) => ix == 0` 

should be added. Also make sure that the index is between the corresponding values

`if ix > 0 and ix < len(a) => a[ix-1]  < x < a[ix]`

tests/mapped_struct.py
klaussfreire: Does assertFalse exist?

tzulberti-jampp: It seems: https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertFalse

It also exists on Python 3

tests/mapped_struct.py
```diff
@@ -591,3 +593,43 @@ class MappedStringMappingBigTest(MappedStringMappingTest):
...
+            a = numpy.array(testarray[:50], dtype)
+            b = numpy.array([0] + testarray[50:] + range(101,103), dtype)
+            a = numpy.sort(a)
+            self.assertEqual(mapped_struct.bsearch(a, 0), 0)
+            self.assertEqual(mapped_struct.bsearch(a, 101), len(a))
+            self.assertEqual(mapped_struct.bsearch(a, 102), len(a))
+            for x in a:
+                ix = mapped_struct.bsearch(a, x)
+                self.assertLess(ix, len(a))
+                self.assertEqual(a[ix], x)
```
tzulberti-jampp, line 616: Misising test for the case where the value is more than one time on the array: [1, 2, 2, 3, 4, 5], and that the first index should be returned on that case

klaussfreire: TBH, the interface is aimed at using numpy arrays as sets, so having duplicates shouldn't be a thing. But lets test in anyway.

sharedbuffers/mapped_struct.py
zeitos: If I'm correct the current usage guarantees that hint <hi.

klaussfreire: 1. hinted_bsearch takes a hint and is externally supplied
2. I tend not to remove segfault-preventing checks assuming I'm not going to fuck up in other parts of the code. Remember: cython compiles to C, and C segfaults when you access an array out of bounds.

zeitos: What I'm saying is: this is a private method, used inside the class. you're doing defensive programming on a private method. If it were public you have a point. 
The calls to this method use hint as lo+hi//2 so the only way to get a segmentation fault is if lo and hi are 0, but you have a lo<hi check.

As you said: hinted_bsearch takes the externally supplied value and in that case, and since that method is the public one you should check the min there.

sharedbuffers/mapped_struct.py
```diff
@@ -1760,6 +1823,253 @@ def _c_search_hkey_ui32(hkey, pindex, stride0, length, hint):
...
+    # Commented cython directives in pxd
+    #@cython.ccall
+    @cython.locals(
+        lo = cython.size_t, hi = cython.size_t, hint = cython.size_t, stride0 = cython.size_t,
+        indexbuf = 'Py_buffer', pindex = cython.p_char)
+    #@cython.returns(cython.size_t)
+    def hinted_bsearch(a, hkey, hint):
+        hi = len(a)
+        lo = 0
+        if hkey < a[0]:
```
zeitos: if a is an empty list, this will raise a segmentation fault?

sharedbuffers/mapped_struct.py
klaussfreire: (as we talked) Even the internal calls aren't so clearly within bounds, the arithmetic could overflow and it's not easy to prove that it doesn't. In fact, `(lo+hi)//2` could in fact overflow if `lo > 0`, which is a use case I intend to allow in the near future (ie: to work on slices)

sharedbuffers/mapped_struct.py
klaussfreire: Just added a test. Indeed it's a bug, though it doesn't segfault. `a[0]` is a regular python style getitem on a numpy array, numpy does the bounds check. But it's not the desired result, so I'm fixing it.

tzulberti-jampp (approved):

sharedbuffers/mapped_struct.py
```diff
@@ -1708,6 +1712,65 @@ def _c_search_hkey_ui64(hkey, pindex, stride0, length, hint):
...
+                        lo = mid - skip
+                        break
+            else:
+                # hit, but must find the first
+                # good idea to go sequential because we assume collisions are unlikely
+                while mid > lo and cython.cast(cython.p_longlong, pindex + stride0 * (mid-1))[0] == hkey:
+                    mid -= 1
+                return mid
+        # Final stretch: search the remaining range with a regular binary search
+        while lo < hi:
```
zeitos, line 1759: If you add a hkey!=mkey here wouldn't you be able to do the while afterwards and prevent code duplication?
like:

```
if lo< hi:
    #first iteration
while mkey != hkey && lo<hi
    #bsearch
if mkey == hkey
    while mid > lo and cython.cast....:
         mid -=1
    return mid
return lo
```

klaussfreire: It would add a superfluous comparison on all iterations (in the other code the `==` condition is implied by being `!<` and `!>`). It would probably measurably hit performance.

There's far more duplication in all the copies for the different types. I'd rather attack that first. But it can wait for another PR right?

sharedbuffers/mapped_struct.py
```diff
@@ -1760,6 +1823,255 @@ def _c_search_hkey_ui32(hkey, pindex, stride0, length, hint):
...
+        mid = cython.size_t, mid2 = cython.size_t, stride0 = cython.size_t, hint = cython.size_t,
+        pindex = cython.p_char, skip = cython.size_t)
+    @cython.returns(cython.size_t)
+    def _c_search_hkey_i32(hkey, pindex, stride0, length, hint):
+        hi = length
+        lo = 0
+        ikey = hkey
+        if lo < hi:
+            mid = min(hint, hi-1)
+            imkey = cython.cast(cython.p_int, pindex + stride0 * mid)[0]
```
zeitos, line 1840: if hkey is longlong shouldn't this cast be p_longlong?

sharedbuffers/mapped_struct.py
```diff
@@ -1760,6 +1823,255 @@ def _c_search_hkey_ui32(hkey, pindex, stride0, length, hint):
...
+        hkey = cython.longlong,
+        mkey = cython.longlong, ikey = cython.int, imkey = cython.int,
+        lo = cython.size_t, hi = cython.size_t, length = cython.size_t,
+        mid = cython.size_t, mid2 = cython.size_t, stride0 = cython.size_t, hint = cython.size_t,
+        pindex = cython.p_char, skip = cython.size_t)
+    @cython.returns(cython.size_t)
+    def _c_search_hkey_i32(hkey, pindex, stride0, length, hint):
+        hi = length
+        lo = 0
+        ikey = hkey
```
zeitos, line 1837: in the other methods you used hkey directly, why do you need ikey here?

sharedbuffers/mapped_struct.py
klaussfreire: No. It's the `i32` type, which means the array is an array of 32-bit ints. That means `p_int`.

sharedbuffers/mapped_struct.py
klaussfreire: It's to avoid type casts within the loop, which would be hard on the compiler. Operations on matching types are easier to optimize (ie: they don't have to concern themselves with overflow for instance).

Other methods are uiXX (which use uikey, similar thing), and the float versions, which just use the correct type for hkey and don't need a typecasted local.

sharedbuffers/mapped_struct.py
```diff
@@ -1760,6 +1823,255 @@ def _c_search_hkey_ui32(hkey, pindex, stride0, length, hint):
...
+        lo = 0
+        if lo < hi:
+            # First iteration a quick guess assuming uniform distribution of keys
+            mid = min(hint, hi-1)
+            mkey = cython.cast(cython.p_float, pindex + stride0 * mid)[0]
+            if mkey < hkey:
+                # Got a lo guess, now skip-search forward for a hi
+                lo = mid = mid+1
+                skip = 32
+                while skip > 0 and mid + skip < hi:
```
zeitos, line 1958: what about using another variable `next_mid` = mid+skip so you don't have to do the same addition 3 times?

klaussfreire: I started doing this, but:
1. It touches a lot of code that works. Violates the "don't fix it if it works" rule massively.
2. It's not clearer really. In some parts of the code it's not clearly a "next_mid", so naming the intermediate value gets confusing. Especially when adjusting downwards, when care is needed to avoid underflow and it's not always the same expression. Eg: one such line would be `lo = next_mid`, which sounds rather misleading.
3. The GCC compiler can easily optimize out the "mid + skip" or "mid - skip" recurring expressions so it's not a performance issue.

So, having tried it, I don't see this as a net improvement.

klaussfreire: Just for clarity, it means I'm against this change, because the resulting code is less clear than the current code.

zeitos (approved):

PR #2, opened 2017-02-15
Setup requires

klaussfreire:
- Fix setup.py to fail properly when built without Cython and without an explicit
  disable of cython optimizations
- Fix setup.py to try to automatically install Cython as a dependency if not
- Require setuptools >= 20. Earlier versions can't handle Cython.Build


setup.py
```diff
@@ -32,13 +70,56 @@ def __init__(self, *p, **kw):
...
+        def __len__(self):
+            self.initialize()
+            return list.__len__(self)
+        def __nonzero__(self):
+            self.initialize()
+            return len(self) > 0
+    extra['ext_modules'] = lazy_modules()
+    extra['setup_requires'] = setup_requires
+
+VERSION = "0.3.4"
```
tzulberti-jampp, line 121: It is missing the version bump on `sharedbuffers/__init__.py`

setup.py
```diff
@@ -32,13 +70,56 @@ def __init__(self, *p, **kw):
...
+    # remove it from setup_requires to keep it in install_requires
+    setup_requires = []
+elif cythonize is None:
+    from distutils.command.build import build
+    class MissingCython(build):
+        def run(self):
+            pkg_resources.require('Cython')
+    cmd_class.update({
+        'build' : MissingCython,
+        'install' : MissingCython,
```
tzulberti-jampp, line 89: MissingCython only extends the build command. But on the install it will use the build. How does that works?

klaussfreire: MissingCython ends up erroring out, so it doesn't really matter.

We know we're missing cython, and run() requires cython. So it's just a fancy way of throwing the right kind of error.

tzulberti-jampp (approved):

zeitos (approved):

PR #6, opened 2017-06-22
Embedded schemas

klaussfreire:
Allow schemas to be picklable, and embed a pickled schema inside
MappedArrays, to make mapping safe across client code versions


sharedbuffers/mapped_struct.py
```diff
@@ -1112,20 +1112,43 @@ def Proxy(self):
...
+        return dict(
+            slot_types = self.slot_types,
+            slot_keys = self.slot_keys,
+            alignment = self.alignment,
+        )
+
+    def __setstate__(self, state):
+        # These are a safety hazard (they could cause OOM if corrupted), so ignore them if present
+        # Not needed for unpacking anyway
+        state.pop('pack_buffer_size', None)
```
tzulberti-jampp, line 1133: I am not sure about this... If during the `__init__.py` it uses a different value than the default, then that value is going to be lost because the `init` uses the default method.

sharedbuffers/mapped_struct.py
```diff
@@ -1547,6 +1588,22 @@ def __init__(self, buf, offset = 0, idmap = None, idmap_size = 1024):
...
+                    raise ValueError("Corrupted input - bad schema location")
+                stored_schema = cPickle.loads(bytes(buffer(buf, self.schema_offset, self.schema_size)))
+                if not isinstance(stored_schema, Schema):
+                    raise ValueError("Corrupted input - unrecognizable schema")
+                if self.schema is None or not self.schema.compatible(stored_schema):
+                    self.schema = stored_schema
+            elif self.schema is None:
+                raise ValueError("Cannot map schema-less buffer without specifying schema")
+        else:
+            self.version = 0
```
tzulberti-jampp: I couldn't find were the `version` attribute is used.

sharedbuffers/mapped_struct.py
klaussfreire: Well, that's the whole point. Pack buffer size only matters for packing, we want to pickle schemas for unpacking purposes only, and also pack buffer size controls memory allocation directly - we don't want external input controlling memory allocation like that.

sharedbuffers/mapped_struct.py
klaussfreire: It's not. I just thought it might be interesting some day to know.

klaussfreire: You know... on-disk format alterations aren't simple, so better add now what we think might be useful later.

tzulberti-jampp (approved):

sharedbuffers/mapped_struct.py
```diff
@@ -1112,20 +1112,43 @@ def Proxy(self):
...
+        if not isinstance(other, Schema):
+            return False
+
+        other_schema = other
+        if self.slot_keys != other_schema.slot_keys or self.alignment != other_schema.alignment:
+            return False
+
+        for k in self.slot_keys:
+            if self.slot_types[k] is not other_schema.slot_types[k]:
+                return False
```
zeitos, line 1149: Consider using all
```
all(k in other_schema.slot_types and other_schema.slot_types[k] == v for k, v in self.slotkeys.items())
```

sharedbuffers/mapped_struct.py
zeitos: you mean know it from previous versions? ... how would that work here? b/c that makes sense for logging extra values but I'm not sure there's much value in keeping unused code here.

zeitos: lets talk this in person.

sharedbuffers/mapped_struct.py
```diff
@@ -1607,14 +1664,16 @@ def __len__(self):
...
+    @cython.locals(schema = Schema, data_pos = cython.size_t, initial_pos = cython.size_t, current_pos = object,
+        schema_pos = cython.size_t, schema_end = cython.size_t)
     def build(cls, initializer, destfile = None, tempdir = None, idmap = None):
         if destfile is None:
             destfile = tempfile.NamedTemporaryFile(dir = tempdir)
 
         initial_pos = destfile.tell()
         write = destfile.write
         write(cls._Header.pack(0, 0, 0))
+        write(cls._NewHeader.pack(cls._CURRENT_VERSION, 0, 0))
```
zeitos: What happens if it's an older or newer version?

sharedbuffers/mapped_struct.py
```diff
@@ -1642,9 +1701,15 @@ def build(cls, initializer, destfile = None, tempdir = None, idmap = None):
...
         destfile.flush()
+
+        schema_pos = destfile.tell()
+        cPickle.dump(schema, destfile, 2)
+        destfile.flush()
+
         final_pos = destfile.tell()
         destfile.seek(initial_pos)
         write(cls._Header.pack(final_pos - initial_pos, index_pos - initial_pos, len(index)))
+        write(cls._NewHeader.pack(cls._CURRENT_VERSION, schema_pos - initial_pos, final_pos - schema_pos))
```
zeitos: same here...

zeitos (changes requested):

sharedbuffers/mapped_struct.py
klaussfreire: Using all there is complex. All takes a generator/list, and that slot_types there is a private attribute of the Schema only accessible through Cython. I'm pretty sure to make it work it would either be ugly (making a type-annotated generator that would take 4x the line count) or inefficient (building a list)

sharedbuffers/mapped_struct.py
klaussfreire: We don't write any other versions than the current

sharedbuffers/mapped_struct.py
klaussfreire: Same again, while writing, we only care about the current version

sharedbuffers/mapped_struct.py
klaussfreire: I was talking about the version field in the on-disk format.

The attribute also makes sense, to have it for future code if it needs to know the version somehow. Note that future code could be external code. So it's an interface added because it seems useful, even if we make no immediate use of it.

sharedbuffers/mapped_struct.py
zeitos: but what happen when you call pack with an older version?

zeitos (approved):

PR #7, opened 2017-06-28
Implement efficient ID mapper merge operations

klaussfreire:
Also, fix pure-python mode to actually work, implement unimplemented
pure-python paths, and adapt tests


sharedbuffers/mapped_struct.py
```diff
@@ -2059,33 +2064,433 @@ def hinted_bsearch(a, hkey, hint):
...
-        hi = len(a)
-        lo = 0
-        return hinted_bsearch(a, hkey, (lo+hi)//2)
+    def _c_merge_ui64(pindex1, length1, pindex2, length2, pdest, destlength, stride0):
+        # Main merge
+        pend1 = pindex1 + stride0 * length1
+        pend2 = pindex2 + stride0 * length2
+        pdestend = pdest + stride0 * destlength
+        pdeststart = pdest
+        while pindex1 < pend1 and pindex2 < pend2 and pdest < pdestend:
```
zeitos, line 2190: `pdestart < pdestend` seems more readable

sharedbuffers/mapped_struct.py
```diff
@@ -2059,33 +2064,433 @@ def hinted_bsearch(a, hkey, hint):
...
-    #@cython.returns(cython.bint)
-    def sorted_contains(a, hkey):
-        hi = len(a)
-        lo = 0
-        return hinted_sorted_contains(a, hkey, (lo+hi)//2)
+        #lint:disable
+        PyObject_GetBuffer(index1, cython.address(index1buf), PyBUF_STRIDED_RO)
+        try:
+            PyObject_GetBuffer(index2, cython.address(index2buf), PyBUF_STRIDED_RO)
+            try:
```
zeitos, line 2450: there's something weird in these 2 trys, also consider using only one try and check if the buffer was initialized in the finally

sharedbuffers/mapped_struct.py
```diff
@@ -2461,6 +2874,61 @@ def map_file(cls, fileobj, offset = 0, size = None):
...
+        indexpos = basepos + cls._Header.size
+
+        # Merge the indexes
+        def merge_all(parts):
+            if len(parts) == 1:
+                return parts[0]
+            else:
+                nparts = []
+                for i in xrange(0, len(parts), 2):
+                    if i+1 < len(parts):
```
zeitos: I this it'll be more readabot to check do :
```
if len(parts)%2 != 0:
     nparts.append(parts[-1])
```
after the loop. you'll also improve performance by doing just one "if"

sharedbuffers/mapped_struct.py
```diff
@@ -2461,6 +2874,61 @@ def map_file(cls, fileobj, offset = 0, size = None):
...
+                        npart = numpy.empty((len(parts[i])+len(parts[i+1]), 2), dtype)
```
zeitos: wouldn't it be better to eliminate the clausure by passing dtype?

sharedbuffers/mapped_struct.py
```diff
@@ -2461,6 +2874,61 @@ def map_file(cls, fileobj, offset = 0, size = None):
...
+                        merge_elements = index_merge(parts[i], parts[i+1], npart)
+                        if merge_elements != len(npart):
+                            npart = npart[:merge_elements]
+                        nparts.append(npart)
+                    else:
+                        nparts.append(parts[i])
+                del parts
+                return merge_all(nparts)
+        index = merge_all([mapper.index for mapper in parts])
```
zeitos: index is a weird name for something that supports `len` operation, what's that index?

sharedbuffers/mapped_struct.py
```diff
@@ -2461,6 +2874,61 @@ def map_file(cls, fileobj, offset = 0, size = None):
...
+
+        write(buffer(index))
+        nitems = len(index)
+
+        finalpos = destfile.tell()
+        if finalpos & 31:
```
zeitos, line 2997: I personally think that when doing binary operations it's better to write `31` as `0x1f`

sharedbuffers/mapped_struct.py
```diff
@@ -2825,7 +3301,10 @@ def build(cls, initializer, destfile = None, tempdir = None):
         try:
             dtypemax = ~dtype(0)
         except:
-            dtypemax = ~0
+            try:
+                dtypemax = ~dtype.type(0)
+            except:
+                dtypemax = ~0
```
zeitos, line 3385: so this is -1 ?

sharedbuffers/mapped_struct.py
```diff
@@ -2825,7 +3301,10 @@ def build(cls, initializer, destfile = None, tempdir = None):
         try:
             dtypemax = ~dtype(0)
         except:
-            dtypemax = ~0
+            try:
```
zeitos, line 3382: I don't understand how dtype can throw an exception, who sets it to None?

zeitos (changes requested):
Great to see this moving forward, I have some questions

sharedbuffers/mapped_struct.py
klaussfreire: It would be incorrect. pdest moves, pdeststart doesn't

sharedbuffers/mapped_struct.py
klaussfreire: Nope, each try is independent, if the second or third initialization fail, the ones that succeeded have to be properly released, that's why there are nested try/finally blocks. They cannot be unified without a lot of spaghetti. Error-prone spaghetti.

sharedbuffers/mapped_struct.py
klaussfreire: Comment doesn't parse. Check syntax.

sharedbuffers/mapped_struct.py
klaussfreire: Possibly

sharedbuffers/mapped_struct.py
klaussfreire: It's index as in the index of a book, a data structure, a thing holding pointers to the elements.

sharedbuffers/mapped_struct.py
klaussfreire:
```
>>> import numpy
>>> a = numpy.array([1])
>>> a.dtype
dtype('int64')
>>> dtype = a.dtype
>>> dtype(0)
TypeError: 'numpy.dtype' object is not callable
>>> 
```

It depends on where the dtype comes from. When using the ones at module-level, it works:

```
>>> dtype = numpy.uint64
>>> ~dtype(0)
18446744073709551615
```

So that except is just in case we get another dtype (the attribute can be overridden by subclasses)

sharedbuffers/mapped_struct.py
klaussfreire: No. The dtype is usually unsigned.

sharedbuffers/mapped_struct.py
zeitos: got it

sharedbuffers/mapped_struct.py
zeitos:
```
if len(parts)%2 != 0:
     nparts.append(parts[:-1])
```

sharedbuffers/mapped_struct.py
zeitos: then instead of an exception it might format your hard drive?

sharedbuffers/mapped_struct.py
zeitos: but 0 is usually 0

klaussfreire: But ~0 isn't usually 0, nor is it always -1

sharedbuffers/mapped_struct.py
klaussfreire: Sure, but we're all grownups here

sharedbuffers/mapped_struct.py
zeitos: @klaussfreire ping

sharedbuffers/mapped_struct.py
zeitos: please explain

sharedbuffers/mapped_struct.py
klaussfreire: I have no idea what you're saying here

sharedbuffers/mapped_struct.py
klaussfreire: For uint32, dtypemax = 0xFFFFFFFF

For uint64, dtypemax = 0xFFFFFFFFFFFFFFFF

zeitos: got it

zeitos (approved):

sharedbuffers/mapped_struct.py
```diff
@@ -2059,33 +2064,451 @@ def hinted_bsearch(a, hkey, hint):
                 raise NotImplementedError("Unsupported array type %s" % (chr(dtype),))
         finally:
             PyBuffer_Release(cython.address(indexbuf)) #lint:ok
+else:
+    import bisect
+    def _hinted_bsearch(a, hkey, hint):
+        return bisect.bisect_left(a, hkey)
+    globals()['hinted_bsearch'] = _hinted_bsearch
+
+#@cython.ccall
```
tzulberti-jampp, line 2149: Should this commented decorators be removed or they are used when running tests or something like that?

klaussfreire: They're informative. The declaration conflicts (ie: Cython doesn't let it be) with the one in the pxd. The pxd rules. But the pxd is invisible, so I left them commented for readers to know that this, is a ccall function. Without having to go to the pxd.

tzulberti-jampp (approved):

PR #10, opened 2017-07-18
Fix complex schema pickling

klaussfreire:
Fix the case of complex schema unpickling, where typed object
references would not register the unpickled shcema with mapped_object
and would end up throwing KeyError


tzulberti-jampp (approved):

tzulberti-jampp (commented):
Missing version bump on setup.py and _version.py

zeitos (approved):

Issue #12, opened 2017-08-02
High peak memory usage building id mappers


klaussfreire:
Id mappers use several times their final size when being built. That seems a bit excessive.

PR #13, opened 2017-08-02
ISSUE-12: Reduce peak memory usage when building

klaussfreire:
Should improve the situation in issue #12 : 
* Concatenate without building a temporary numpy array duplicating the contents
* Dereference temporary arrays earlier
* Coalesce big parts list into smaller lists of big parts to reduce memory fragmentation and improve deduplication
* Specialize the algorithm used by numpy.unique for our purposes, to avoid the RAM wastage induced by numpy's generality. In particular:
  - Don't use argsort, the indices array weighs a lot, use in-place sort (since we feed it scratch buffers, we can alter the input)
  - Compute flags using sliced views of the input array
  - Don't copy the array if all flags are true, just return it
  - Don't compute indexes out of the flags, just use the flags


sharedbuffers/mapped_struct.py
```diff
@@ -2622,6 +2623,21 @@ def _merge_all(parts, dtype):
         del parts
         return _merge_all(nparts, dtype)
 
+@cython.cfunc
+@cython.locals(parts = list, discard_duplicate_keys = cython.bint, discard_duplicates = cython.bint)
+def _discard_duplicates(apart, struct_dt, discard_duplicate_keys, discard_duplicates):
+    if discard_duplicate_keys or discard_duplicates:
+        # What numpy.unique does, but using MUCH less RAM
+        vpart = apart.view(struct_dt)
+        vpart.sort(0)
```
zeitos, line 2632: what's 0?

sharedbuffers/mapped_struct.py
```diff
@@ -2622,6 +2623,21 @@ def _merge_all(parts, dtype):
         del parts
         return _merge_all(nparts, dtype)
 
+@cython.cfunc
+@cython.locals(parts = list, discard_duplicate_keys = cython.bint, discard_duplicates = cython.bint)
+def _discard_duplicates(apart, struct_dt, discard_duplicate_keys, discard_duplicates):
+    if discard_duplicate_keys or discard_duplicates:
+        # What numpy.unique does, but using MUCH less RAM
+        vpart = apart.view(struct_dt)
```
zeitos, line 2631: where can I find the documentation for view?

klaussfreire: https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.view.html

sharedbuffers/mapped_struct.py
klaussfreire: The first axis (first dimension)

zeitos: talked in person

zeitos (approved):

tzulberti-jampp (approved):

PR #16, opened 2017-08-24
Unpack frozensets as frozensets

klaussfreire:
Instead of unpacking them as subclasses of frozensets.
This matches other primitive types, is more interoperable with
code that expects frozensets, and prevents breaking singleton
properties of the frozensets (that empty frozensets unpack
as a single instance no matter where)


tzulberti-jampp (approved):

klaussfreire:
Ping @GervasioCalderon @zeitos

sharedbuffers/_version.py
```diff
@@ -1 +1 @@
-__version__ = '0.4.2'
\ No newline at end of file
+__version__ = '0.4.3'
```
zeitos: shouldn't this file have a new line?

klaussfreire: It's autogenerated by setup.py, if it's missing we can make the change in setup.py to add the newline.

zeitos: ok

zeitos (approved):

PR #17, opened 2017-08-25
Script to run tests on docker

torian:
Little script to run tests quickly and easily on docker by
excecuting:

```
 $> docker run --rm \
  -v ${PWD}:/opt/sharedbuffers \
  -w /opt/sharedbuffers \
  python:2.7 \
    /bin/sh run-tests.sh
```

Can also be used to run tests locally through virtualenv:

```
 $> sh ./run-tests.sh virtualenv
```


sebamontini (approved):
LGTM

zeitos (approved):

tzulberti-jampp (approved):

README.rst
```diff
@@ -156,3 +156,13 @@ rather than the first one, so multi-maps and simple mappings are binary compatib
...
 returned values could have extra values corresponding to other keys whose hash collide with the one being requested.
+
+Tests
+=====
+
+Tests can be run on docker with the following command line:
+
+.. code::shell
+
+  $> docker run -v ${PWD}:/opt/sharedbuffers -w /opt/sharedbuffers python:2.7 /bin/sh run-docker-tests.sh
```
klaussfreire: Docker is absolutely NOT the recommended way of running tests, so I wouldn't document this as the only way to do it.

If there's only one way to run tests documented, that's the way people will think is *the* way. This is not it.

Besides, that script there has nothing to do with docker. That's just a script to install requirements and run tests. You might as well drop the mention of docker and it would still be accurate.

run-docker-tests.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+set -e
+
+PATH=/usr/local/bin:$PATH
```
klaussfreire: There's no reason/need to override the path

run-docker-tests.sh
```diff
@@ -0,0 +1,12 @@
...
+
+python --version
+
+pip install -q -r requirements.txt
```
klaussfreire: Why quiet?

klaussfreire (changes requested):

README.rst
torian: Fair enough. I'll update the documentation.

run-docker-tests.sh
torian: The idea is to run the python version that the image ships, which is located in `/usr/local/bin`, otherwise, it would run the tests with an older version (`/usr/bin`).

I can change that to an absolute path, if preferred.

run-docker-tests.sh
torian: just to keep the interesting output, and not irrelevant things that pip might want to spit out.

klaussfreire: The versions of installed requirements seem quite interesting to me

run-docker-tests.sh
klaussfreire: It seems `/usr/local/bin` [is there by default](https://docs.docker.com/engine/reference/run/#env-environment-variables)

klaussfreire: In any case, this script has no knowledge of the image being run, so any PATH manipulations should be done on the `docker run` command in the doc, that does know.

README.rst
```diff
@@ -156,3 +156,20 @@ rather than the first one, so multi-maps and simple mappings are binary compatib
 Multi-maps with string keys can also be approximate, meaning the original keys will be discarded and the mapping will
 only work with hashes, making the map much faster and more compact, at the expense of some inaccuracy where the
 returned values could have extra values corresponding to other keys whose hash collide with the one being requested.
+
+Tests
+=====
+
+Running tests can be done locally or on docker, using the script `run-tests.sh`:
```
klaussfreire, line 163: Include the virtualenv steps you had included earlier, that were a good idea.

```
Running tests can be done locally in a virtual env, or alternatively in a docker, using the script `run-tests.sh`:

.. code:: shell

  $> virtualenv venv
  $> . ./venv/bin/activate
  $> ./run-tests.sh
```

Make sure run-tests.sh has execute permission, btw.

klaussfreire (changes requested):

README.rst
klaussfreire: After that, LGTM

klaussfreire (approved):

PR #20, opened 2017-10-03
Small optimization in MHM build

klaussfreire:
Avoid some unnecessary work in _discard_duplicates when there
are no duplicates


sharedbuffers/mapped_struct.py
```diff
@@ -2624,17 +2624,18 @@ def _merge_all(parts, dtype):
...
         # What numpy.unique does, but using MUCH less RAM
         vpart = apart.view(struct_dt)
         vpart.sort(0)
         if discard_duplicate_keys:
-            flags = numpy.concatenate([[True], apart[1:,0] != apart[:-1,0]])
+            flags = apart[1:,0] != apart[:-1,0]
         elif discard_duplicates:
-            flags = numpy.concatenate([[True], vpart[1:,0] != vpart[:-1,0]])
+            flags = vpart[1:,0] != vpart[:-1,0]
         if not numpy.all(flags):
```
zeitos: no need to call all now right?

klaussfreire: Why not?

zeitos: my bad, != is overriden

sharedbuffers/mapped_struct.py
```diff
@@ -2624,17 +2624,18 @@ def _merge_all(parts, dtype):
...
+            flags = numpy.concatenate([[True], flags])
             apart = apart[flags]
```
zeitos: the name is `discard_duplicates`
but:
```
array([[ 1.,  0.,  0.],
       [ 2.,  3.,  4.],
       [ 1.,  2.,  4.]])
>>> apart
array([[ 1.,  0.,  0.],
       [ 1.,  1.,  2.],
       [ 2.,  3.,  4.],
       [ 1.,  2.,  4.],
       [ 1.,  3.,  5.]])
>>> apart[1:,0] != apart[:-1,0]
array([False,  True,  True, False], dtype=bool)
>>> apart[[numpy.concatenate([[True],(apart[1:,0] != apart[:-1,0])])]]
array([[ 1.,  0.,  0.],
       [ 2.,  3.,  4.],
       [ 1.,  2.,  4.]])
```

klaussfreire: You're using the `discard_duplicate_keys` code path, which is supposed to deduplicate based on the first column only. So it's correct.

Furthermore, you're skipping the sort pass, which makes the result also incorrect.

zeitos (approved):

sharedbuffers/mapped_struct.py
```diff
@@ -2634,9 +2635,33 @@ def _discard_duplicates(apart, struct_dt, discard_duplicate_keys, discard_duplic
...
+            else:
+                # In-place compress operation, best option for non-copying deduplication, to avoid
+                # having to allocate temporary workspace proportional to output size. This
+                # way only uses a fixed amount of temporary workspace and builds a compressed
+                # result into the input array directly.
+                nmin = np.argmin(flags)
+                count_nonzero = np.count_nonzero
+                start = nmin
+                end = len(flags)
+                out_start = start+1
```
zeitos, line 2653: why +1?

klaussfreire: Because of the first implicit True

sharedbuffers/mapped_struct.py
```diff
@@ -2634,9 +2635,33 @@ def _discard_duplicates(apart, struct_dt, discard_duplicate_keys, discard_duplic
...
+                while start < end:
+                    chunk_end = min(end, start + 100000)
+                    flags_slice = flags[start:chunk_end]
+                    nout = count_nonzero(flags_slice)
+                    dedup_slice = apart[start+1:chunk_end+1]
+                    if nout != (chunk_end - start):
```
zeitos: why not `nout < (chunk_end - start)`? it shouldn't be bigger

sharedbuffers/mapped_struct.py
```diff
@@ -2634,9 +2635,33 @@ def _discard_duplicates(apart, struct_dt, discard_duplicate_keys, discard_duplic
...
+                nmin = np.argmin(flags)
+                count_nonzero = np.count_nonzero
+                start = nmin
+                end = len(flags)
+                out_start = start+1
+                while start < end:
+                    chunk_end = min(end, start + 100000)
+                    flags_slice = flags[start:chunk_end]
+                    nout = count_nonzero(flags_slice)
+                    dedup_slice = apart[start+1:chunk_end+1]
```
zeitos: I don't understand why the +1s here :S

sharedbuffers/mapped_struct.py
klaussfreire: If it's bigger it's a bug, skipping the operation would silence that bug, carrying it out would make numpy raise an exception and expose it.

sharedbuffers/mapped_struct.py
klaussfreire: Because the flags array is missing the initial True, this compensates for that implicit True

sharedbuffers/mapped_struct.py
zeitos: if that's the case wouldn't it be better to have our own exception that explains the reason?
```
else:
   if nout > (chunk_end - start):
       raise ValueError(("Since python can't create matter, "
            "there's got to be something really wrong. We found "
            "more non zeros than the array size! some context info:"
            " end: %s, start: %s, nout: %s, slice: %s"), end, start, nout, flags_slice)
```

zeitos (changes requested):

sharedbuffers/mapped_struct.py
klaussfreire: No, numpy's exception is very precise and informative

klaussfreire: I'm going to cry timeout here...

sharedbuffers/mapped_struct.py
```diff
@@ -2624,18 +2624,44 @@ def _merge_all(parts, dtype):
...
+                flags = np.concatenate([[True], flags])
+                apart = apart[flags]
+            else:
+                # In-place compress operation, best option for non-copying deduplication, to avoid
+                # having to allocate temporary workspace proportional to output size. This
+                # way only uses a fixed amount of temporary workspace and builds a compressed
+                # result into the input array directly.
+                nmin = np.argmin(flags)
+                count_nonzero = np.count_nonzero
+                start = nmin
```
tzulberti-jampp, line 2651: Why can't you do `start = nmin + 1` to prevent doing the `+1` all the time?

klaussfreire: If you read carefully, you'll see that the start/end/chunk_end variables get used in both contexts, those that need the +1 and those that don't.

tzulberti-jampp (approved):

zeitos (approved):

zeitos (approved):

PR #21, opened 2017-12-13
0.27.1 doesn't build

klaussfreire:


tzulberti-jampp (approved):

tzulberti-jampp (commented):
Shouldn't we do a hotfix release with this?

requirements.txt
```diff
@@ -1,6 +1,6 @@
 lz4
 xxhash
 numpy
-Cython>=0.22
+Cython>=0.22,!=0.27.1
```
tzulberti-jampp: I updated to 0.27.3 and still got the following compilation error:

``` 
Error compiling Cython file:
------------------------------------------------------------
...
                                    self.index[startpos,1],
                                    pybuf.len) == bkey:
                                return True
                            startpos += 1
                finally:
                    PyBuffer_Release(cython.address(pybuf))
                                          ^
------------------------------------------------------------

sharedbuffers/mapped_struct.py:4055:43: Taking address of non-lvalue (type Python object)
```

klaussfreire: Seems like a different error. Will look into it on monday

tzulberti-jampp: Ok, I will create a new ticket for this error: https://github.com/jampp/sharedbuffers/issues/23

klaussfreire: It seems like another bug in Cython

tzulberti-jampp (approved):

pablodo (approved):

zeitos (approved):

Issue #24, opened 2018-01-03
Error when using lz4 version 0.18.1


tzulberti-jampp:
When using lz4 0.18.1 the following error take place:

```
  File "multihashmap.py", line 6, in <module>
    import sharedbuffers
  File "sharedbuffers/__init__.py", line 4, in <module>
    from .mapped_struct import *
  File "sharedbuffers/mapped_struct.py", line 301, in init sharedbuffers.mapped_struct (sharedbuffers/mapped_struct.c:90845)
    lz4_decompress = cython.declare(object, lz4.decompress)
AttributeError: 'module' object has no attribute 'decompress'
```

Based on the documentation it should use the `frame` module to access that functions: http://python-lz4.readthedocs.io/en/stable/lz4.frame.html

klaussfreire:
> Based on the documentation it should use the `frame` module to access that functions: http://python-lz4.readthedocs.io/en/stable/lz4.frame.html

There, it says:

> This sub-package is in beta testing.

What a great project huh? Deprecating a stable API in favor of a beta. Wow.

PR #25, opened 2018-01-03
Issue 24 lz4 frame module

tzulberti-jampp:
Solve #24


klaussfreire (changes requested):
Nope. They're not interchangeable. A change like this needs far more care and at least a minor bump. Also a version change in the file format, and compatibility code when reading the old format.

tzulberti-jampp:
If I understand correctly the version bump is required because what the file created with lz4 0.8.1 is going to be imcompatible with the one created with version 0.18.1. Not because of the difference on code...

I googled who to make lz4 compatible between versions but I wasn't able to find any information about this. Do you have an idea on who to make those versions compatible?

klaussfreire:
Yes. Use the file format version embedded in the bundles and have both versions of the code.

klaussfreire:
It will probably be more efficient to talk about it in person tomorrow

sharedbuffers/mapped_struct.py
```diff
@@ -298,8 +297,15 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
             raise ValueError("Inconsistent data, unknown type code %r" % (dcode,))
         return rv
 
-lz4_decompress = cython.declare(object, lz4.decompress)
-lz4_compress = cython.declare(object, lz4.compress)
+try:
+    import lz4.block
+    lz4_functions_module = lz4.block
```
klaussfreire: lz4_block?

klaussfreire (changes requested):

klaussfreire (approved):

zeitos (approved):
LGTM

PR #36, opened 2018-08-09
Support proxied container types

gefarion:


sharedbuffers/mapped_struct.py
```diff
@@ -72,6 +72,14 @@ def _likebuffer(buf):
     else:
         return buffer(buf)
 
+@cython.inline
+@cython.cfunc
+def _likerobuffer(buf):
+    if type(buf) is buffer or type(buf) is bytes or isinstance(buf, bytes):
```
klaussfreire: A little docstring explaining mainly the difference with likebuffer wouldn't be a bad idea

sharedbuffers/mapped_struct.py
```diff
@@ -297,6 +306,226 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
...
 
+@cython.cclass
+class proxied_list(object):
+    cython.declare(
+        buf = object,
+        pybuf = 'Py_buffer',
+        offs = cython.ulonglong,
+        dataoffs = cython.ulonglong,
+        itemsize = cython.ulonglong,
+        objlen = cython.ulonglong,
```
klaussfreire: I don't like all the state the proxy has.

For one, it makes them big. For another, it makes them sensitive to buffer changes. We don't want to paint ourselves into the read-only buffer corner, we will implement writable buffers eventually, so the proxy should only point to the list header, and things like the dcode, element size and all that should be read from the header at item access time. If properly cythonized, the overhead shouldn't be big.

sharedbuffers/mapped_struct.py
```diff
@@ -297,6 +306,226 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
...
+        if idmap is None:
+            idmap = {}
+        if offs in idmap:
+            return idmap[offs]
+
+        buf = _likerobuffer(buf)
+        dcode = buf[offs]
+        dataoffs = offs
+
+        if dcode in ('B','b','H','h','I','i'):
```
klaussfreire: Check the code generated here.

Ideally, you'll want a switch. You may have to use `@cython.use_switch(True)` and make sure the `dcode` is a char.

sharedbuffers/mapped_struct.py
```diff
@@ -297,6 +306,226 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
...
+            objlen, = struct.unpack('<I', buf[dataoffs+1:dataoffs+4] + '\x00')
```
klaussfreire: Cython shouldn't be using struct.unpack. Check how other complex structures handle cythonized parsing of the buffer headers and try to build an optimized path here. This is what will happen each time you do:

``` python
some_object.some_list_attribute[3]
```

So it needs to be fast

sharedbuffers/mapped_struct.py
```diff
@@ -297,6 +306,226 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
...
+
+        dcode = self.dcode
+        if dcode == 't':
+            if cython.compiled:
+                obj_offs = self.offs + self.index[index]
+            else:
+                index_offs = self.dataoffs + self.itemsize * index
+                obj_offs = self.offs + self.struct.unpack(self.buf[index_offs:index_offs + self.itemsize])[0]
+        else:
+            obj_offs = self.dataoffs + self.itemsize * index
```
klaussfreire: Why are you doing this instead of simply:

``` python
cast(p_uint, cast(p_char, self.pybuf.buf) + self.dataoffs)[index]
```

sharedbuffers/mapped_struct.py
```diff
@@ -297,6 +306,226 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
...
+
+    @cython.locals(a = 'proxied_list')
+    @staticmethod
+    def _eq(a, b):
+        if a.objlen != len(b):
+            return False
+        for i in xrange(a.objlen):
+            if a[i] != b[i]:
+                return False
+        return True
```
klaussfreire: If it's only for testing that you're implementing `__eq__`, you could simply use generic code here:

``` python
alen = len(a)
if alen != len(b):
    return False
for i in xrange(alen):
    if a[i] != b[i]:
        return False
return True
```

If you want it to be fast, you should have a fast path for the case where both are proxied lists.

You should also not fail in cases where b doesn't support len() or index access (ie: it's another kind of type). You'll need to be safe here if you want to reuse this for mapped tuples (which you probably want), since tuples need to be hashable *and* compare to anything without failing. In fact, lists also compare to anything without failing, so you probably also want that behavior for mapped lists even though I'd agree it'd be best to raise a type error, but python lists don't, so better be consistent with the behavior of the type you're trying to emulate.

You'll also need a full rich compare method (__gt__, etc) for tuples.

sharedbuffers/mapped_struct.py
```diff
@@ -297,6 +306,226 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
...
+
+    @cython.locals(a = 'proxied_tuple')
+    @staticmethod
+    def _eq(a, b):
+        if a.objlen != len(b):
+            return False
+        for i in xrange(a.objlen):
+            if a[i] != b[i]:
+                return False
+        return True
```
klaussfreire: Why do you reimplement _eq if it's the same?

sharedbuffers/mapped_struct.py
```diff
@@ -522,10 +751,12 @@ class mapped_object(object):
         mapped_frozenset : 'Z',
         mapped_tuple : 't',
         mapped_list : 'e',
-        mapped_bytes : 's',
```
klaussfreire, line 525: Where did mapped_bytes go?

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
```diff
@@ -753,6 +764,7 @@ class mapped_object(object):
         mapped_list : 'e',
         mapped_unicode : 'u',
         mapped_bytes : 's',
+        mapped_dict : 'c',
```
klaussfreire: c?

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
```diff
@@ -75,6 +78,9 @@ def _likebuffer(buf):
 @cython.inline
 @cython.cfunc
 def _likerobuffer(buf):
+    """
+    Takes a container as parameter and returns a read-only objecto with buffer protocol.
```
klaussfreire: objecto?

Container? Not really. An object that supports the buffer interface, it's not at all the same.

sharedbuffers/mapped_struct.py
```diff
@@ -317,113 +323,331 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+    def __init__(self, buf, offs):
+        self.buf = buf
+        self.offs = offs
+        self.pos = 0
+
+    def write(self, value):
+        size = len(value)
+
+        offs = self.offs + self.pos
+        self.buf[offs:offs + size] = value
```
klaussfreire, line 339: Will this extend bytearrays when writing beyond the end?

sharedbuffers/mapped_struct.py
```diff
@@ -317,113 +323,331 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+
+    def seek(self, pos):
+        self.pos = pos
+
+@cython.cclass
+class proxied_dict(object):
+
+    cython.declare(
+        value_array = proxied_list,
+        id_mapper   = object,
```
klaussfreire: Don't align

sharedbuffers/mapped_struct.py
```diff
@@ -317,113 +323,331 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+        return self.pos
+
+    def flush(self):
+        pass
+
+    def seek(self, pos):
+        self.pos = pos
+
+@cython.cclass
+class proxied_dict(object):
```
klaussfreire, line 354: What about some docstrings?

Especially about the difference between this and an actual dict. I'll give you one that you might not have noticed: the iteration order between keys, values and items isn't consistent. While in dicts `zip(iterkeys, itervalues) == items()`, it won't be true for this implementation.

sharedbuffers/mapped_struct.py
```diff
@@ -317,113 +323,331 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+        else:
+            return self.value_array[ix]
+
+    def __eq__(self, other):
+
+        if len(self) != len(other):
+            return False
+
+        for k in self:
+            if k not in other or other[k] != self[k]:
```
klaussfreire: This does the lookup twice. You should use get.

sharedbuffers/mapped_struct.py
```diff
@@ -317,113 +323,331 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+
+        all_strs = True
+        all_ints = True
+
+        for k in obj:
+            if type(k) is not str:
+                all_strs = False
+                break
+
+        for k in obj:
```
klaussfreire: You can probably use a single loop

sharedbuffers/mapped_struct.py
```diff
@@ -317,113 +323,331 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+        return len(self.value_array)
+
+    @classmethod
+    def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0):
+
+        all_strs = True
+        all_ints = True
+
+        for k in obj:
+            if type(k) is not str:
```
klaussfreire: I think you're forgetting unicode

sharedbuffers/mapped_struct.py
```diff
@@ -317,113 +323,331 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+            else:
+                ktype = 'S'
+                mapper_class = StringIdMapper
+        elif all_ints:
+            if len(obj) <= 0xFFFFFFFF:
+                ktype = 'i'
+                mapper_class = NumericId32Mapper
+            else:
+                ktype = 'I'
+                mapper_class = NumericIdMapper
```
klaussfreire, line 449: Don't you need special subclasses with a proper hash function?

Numeric will expect all the keys to fit in the integer type you selected, which happens to be both unsigned and narrow if you end up using the 32-bit type. That the mapping contains less than 2³² entries doesn't mean the keys will fit in 32 bits, not to mention unsigned.

Instead of going through all the trouble of checking the key space, you might as well simply hash the keys into the required key space.

sharedbuffers/mapped_struct.py
```diff
@@ -317,113 +323,331 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+        header_offs = offs
+        # Reserve space for the header
+        offs += cls._Header.size
+
+        # We need simulate the io protocol over the buffer to reuse the mapper code
+        iobuf = BufferIO(buf, offs)
+        offs += mapper_class.build(_iter_keys(obj), iobuf, return_map=False)
+
+        value_offs = offs
+        offs += proxied_list.pack_into(obj.values(), buf, offs, idmap, implicit_offs)
```
klaussfreire, line 462: Beware, if this is given a proxied_dict, .keys() iteration order won't match values().

sharedbuffers/mapped_struct.py
```diff
@@ -317,113 +323,331 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+
+        if selfe < othere:
+            return -1
+        elif selfe > othere:
+            return 1
+
+    return 0
+
+def islist(obj):
+    return isinstance(obj, (tuple, list, proxied_list, proxied_tuple))
```
klaussfreire, line 516: You can register proxied_list and other proxied types into collections ABCs, and then you can check `(tuple, list, Sequence)`

sharedbuffers/mapped_struct.py
```diff
@@ -317,113 +323,331 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+        othere = b[i]
+
+        if selfe < othere:
+            return -1
+        elif selfe > othere:
+            return 1
+
+    return 0
+
+def islist(obj):
```
klaussfreire, line 515: ccall and returns(bint)?

sharedbuffers/mapped_struct.py
```diff
@@ -317,113 +323,331 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+            return dcode, objlen, itemsize, dataoffs
 
-    @classmethod
-    def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0):
-        # Same format as tuple, only different base type
-        return mapped_tuple.pack_into(obj, buf, offs, idmap, implicit_offs)
+        elif dcode in ('q', 'd', 't'):
+            objlen = cython.cast(cython.p_longlong, pbuf + dataoffs)[0] >> 8
+            dataoffs += 8
+            return dcode, objlen, 8, dataoffs
```
klaussfreire: I don't think you should do this.

For one, returning a tuple makes it slow (it has to build the tuple).

But most importantly, explicit pointer arithmetic using itemsize won't be as efficient or safe as just casting the pointer into the right type, and indexing the pointer.

You'll have to write more code and might end up repeating yourself a bit, but I think it will generate far better code if you avoid explicitly computing itemsize. Just have a switch by dcode that casts into the right type.

You can have a helper to get the objlen to avoid repeating that code. Make sure calls to all helpers are properly optimized, especially in item accessor methods, that will likely be quite hot.

sharedbuffers/mapped_struct.py
```diff
@@ -317,113 +323,331 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+        self.offs = offs
+        self.buf = buf
+        self.idmap = {} if idmap is None else idmap
+
+        if cython.compiled:
+            self._cinit()
+            self._cmetadata()
+        else:
+            # Call metadata to check the object
+            self._metadata()
```
klaussfreire, line 612: Why different methods?

Why can't you just have the `if cython.compiled` in _metadata?

sharedbuffers/mapped_struct.py
```diff
@@ -317,113 +323,331 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+
+    @cython.ccall
+    def _cinit(self):
+        self.pybuf.buf = cython.NULL
+        PyObject_GetBuffer(self.buf, cython.address(self.pybuf), PyBUF_SIMPLE)  # lint:ok
```
klaussfreire, line 617: Beware, if you do this, you need a `__dealloc__`

sharedbuffers/mapped_struct.py
```diff
@@ -458,34 +682,102 @@ def __getitem__(self, index):
...
+            raise ValueError("Inconsistent data, unknown type code %r" % (dcode,))
+
+        return objlen
 
     def __len__(self):
-        return self.objlen
+        if cython.compiled:
+            return self._cobjlen()
+        else:
+            return self._objlen()
```
klaussfreire: Really... why do you copy methods around instead of making the implementation inside conditional?

sharedbuffers/mapped_struct.py
```diff
@@ -494,46 +786,32 @@ def __delitem__(self, index):
         raise AttributeError("Proxy objects are read-only")
 
     def __iter__(self):
-        for i in xrange(self.objlen):
+        for i in xrange(len(self)):
             yield self[i]
```
klaussfreire: You probably want a `_get` method or something that you can call in an optimized fashion instead of this generic getitem. Iteration I imagine will be a big use case for proxied containers, so you probably want them to be efficient.

sharedbuffers/mapped_struct.py
```diff
@@ -494,46 +786,32 @@ def __delitem__(self, index):
...
+        for e in self:
+            if e ==  item:
                 return True
         return False
 
+    def __repr__(self):
+        return str(self)
+
     def __str__(self):
         return "proxied_list([%s])" % ",".join([str(x) for x in self])
```
klaussfreire: I'd make `str` mimic the str of a list, and `repr` add the `proxied_list` wrapper to clarify that it's not just a list.

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
gefarion: Yes.

sharedbuffers/mapped_struct.py
gefarion: Because these methods are long and I think that it is much clear in this way.

sharedbuffers/mapped_struct.py
gefarion: I am using __del__ for releasing the buffer. Is it not ok?

klaussfreire: It's not the same really. Check out cython docs about it.

sharedbuffers/mapped_struct.py
```diff
@@ -404,7 +406,8 @@ def __eq__(self, other):
             return False
 
         for k in self:
-            if k not in other or other[k] != self[k]:
+            other.get(k, NOT_FOUND_VALUE)
+            if k is NOT_FOUND_VALUE or other[k] != self[k]:
```
klaussfreire: That doesn't seem right

klaussfreire (changes requested):

PR #38, opened 2018-09-20
Add support for datetime and Decimal

gefarion:


sharedbuffers/mapped_struct.py
```diff
@@ -503,6 +521,62 @@ def unpack_from(cls, buf, offs, idmap = None):
             idmap[offs] = rv
         return rv
 
+class mapped_decimal(Decimal):
+    @classmethod
+    @cython.locals(offs = cython.longlong, implicit_offs = cython.longlong, exponent = cython.longlong, sign = cython.uchar)
+    def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0, packer = struct.Struct('=Bq')):
```
klaussfreire: Usually we place the packer and other Structs as class attributes.

Also... do you really need 64 bits for the exponent? Also those 64 bits just after a byte will have a lot of padding.

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
```diff
@@ -638,6 +718,8 @@ def __init__(self, value = None, typ = None, TYPE_CODES = TYPE_CODES):
     unicode : mapped_unicode,
     bytes : mapped_bytes,
     object : mapped_object,
+    datetime : mapped_datetime,
+    Decimal : mapped_decimal
```
klaussfreire: Missing `,`

Also... you should try to import cdecimal and support it in addition to decimal. Same format, same everything, but it's a different class so it has to be included in the type mappings.

klaussfreire (changes requested):

tests/mapped_struct.py
```diff
@@ -1178,23 +1199,35 @@ def testPack(self):
         unpacked_now = mapped_datetime.unpack_from(buf, 2)
         self.assertEquals(now, unpacked_now)
 
+    def testPackOldDate(self):
+        buf = bytearray(12)
+        now = datetime(1900, 01, 01)
```
klaussfreire: Careful with those octal literals

klaussfreire (approved):

tzulberti-jampp (approved):

PR #40, opened 2018-09-21
Add support for mapping from uncompressed zip files

klaussfreire:


tests/mapped_struct.py
```diff
@@ -593,6 +594,19 @@ def testFileMapping(self):
             mapped = self.MappedArrayClass.map_file(destfile)
             self._checkValues(mapped, mapped)
 
+    def testZipMapping(self):
```
gefarion, line 597: Can you add tests with multiple files inside?

klaussfreire: What's the difference?

tzulberti-jampp (approved):

gefarion (approved):

PR #41, opened 2018-09-24
Add support for mapped dict and proxied list/tuples

gefarion:


sharedbuffers/mapped_struct.py
```diff
@@ -320,6 +320,342 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
             raise ValueError("Inconsistent data, unknown type code %r" % (dcode,))
         return rv
 
+class mapped_dict(dict):
+
+    @classmethod
+    def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0):
+        return mapped_list.pack_into((obj.keys(), [obj[k] for k in obj]), buf, offs, idmap, implicit_offs)
```
klaussfreire: I'd rather build both lists explicitly. Why? Because mapped_object is very pedant about the type of things. If .keys() returns something that's a sequence but not really a list, this will fail.

So, I'd suggest:

``` python
keys = list(obj)
values = [ obj[k] for k in keys ]
return mapped_list.pack_into([keys, values], buf, offs, idmap, implicit_offs)
```

sharedbuffers/mapped_struct.py
```diff
@@ -320,6 +320,342 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
...
+        othere = b[i]
+
+        if selfe < othere:
+            return -1
+        elif selfe > othere:
+            return 1
+
+    return 0
+
+def islist(obj):
```
klaussfreire, line 399: These helpers should probably be ccall and return a native data type.

sharedbuffers/mapped_struct.py
```diff
@@ -320,6 +320,342 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
...
+    return isinstance(obj, (tuple, list, proxied_list, proxied_tuple))
+
+@cython.cclass
+class proxied_list(object):
+
+    cython.declare(
+        buf        = object,
+        pybuf      = 'Py_buffer',
+        offs       = cython.ulonglong,
+        idmap      = dict,
```
klaussfreire: Don't align operators, that's anti-PEP8

sharedbuffers/mapped_struct.py
```diff
@@ -320,6 +320,342 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
...
+    )
+
+    if cython.compiled:
+        def __del__(self):
```
klaussfreire, line 412: Did you check that this work correctly done conditional like this? The Cython compiler sometimes fails to understand dynamism when it comes to data or method definitions, and being a special method this, if Cython doesn't understand it, it won't work correctly (it's not a regular method, it needs to be registered as a finalizer in the type struct).

Also, did you check whether you should use `__del__` or `__dealloc__`? Because Cython's recommended finalizer for C data structures is the latter.

sharedbuffers/mapped_struct.py
```diff
@@ -320,6 +320,342 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
...
+                return dcode, objlen, itemsizes['l'], dataoffs, struct.Struct('l')
+
+            else:
+                raise ValueError("Inconsistent data, unknown type code %r" % (dcode,))
+
+    @cython.locals(offs = cython.ulonglong, idmap = dict)
+    def __init__(self, buf, offs, idmap = None):
+        self.offs = offs
+        self.buf = buf
+        self.idmap = {} if idmap is None else idmap
```
klaussfreire: A persistent idmap for the proxy? That sounds like it will backfire.

sharedbuffers/mapped_struct.py
```diff
@@ -320,6 +320,342 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
...
+                    cython.cast(cython.p_uchar, self.pybuf.buf) + obj_offs)[0]  # lint:ok
+            elif dcode == 'd':
+                res = cython.cast(cython.p_double,
+                    cython.cast(cython.p_uchar, self.pybuf.buf) + obj_offs)[0]  # lint:ok
+            else:
+                raise ValueError("Inconsistent data, unknown type code %r" % (dcode,))
+        else:
+            res = _struct.unpack(self.buf[obj_offs:obj_offs + itemsize])[0]
+
+        self.idmap[obj_offs] = res
```
klaussfreire: I'd skip the idmap for primitive types. It's only useful for object types.

sharedbuffers/mapped_struct.py
```diff
@@ -320,6 +320,342 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
...
+        return res
+
+    if cython.compiled:
+        @cython.locals(index = int)
+        def  __richcmp__(self, other, op):
```
klaussfreire: Again, check that the conditional doesn't confuse Cython

sharedbuffers/mapped_struct.py
```diff
@@ -320,6 +320,342 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
...
+        else:
+            raise ValueError("Inconsistent data, unknown type code %r" % (dcode,))
+
+        return objlen
+
+    def __len__(self):
+        if cython.compiled:
+            return self._cobjlen()
+        else:
+            return self._objlen()
```
klaussfreire: Again, don't have 2 versions of the same method, just have the if cython.compiled inside the method.

sharedbuffers/mapped_struct.py
```diff
@@ -320,6 +320,342 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
...
+
+    def __delitem__(self, index):
+        raise AttributeError("Proxy objects are read-only")
+
+    def __iter__(self):
+        for i in xrange(len(self)):
+            yield self[i]
+
+    def __reversed__(self):
+        for i in reversed(xrange(len(self))):
```
klaussfreire: `xrange(len(self)-1, -1, -1)`

sharedbuffers/mapped_struct.py
```diff
@@ -320,6 +320,342 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
...
+            yield self[i]
+
+    def __contains__(self, item):
+        for e in self:
+            if e ==  item:
```
klaussfreire: Check whitespace

sharedbuffers/mapped_struct.py
```diff
@@ -320,6 +320,342 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
...
+                return True
+        return False
+
+    def __repr__(self):
+        return str(self)
+
+    def __str__(self):
+        return "proxied_list([%s])" % ",".join([str(x) for x in self])
```
klaussfreire: I think the str would be just `[...]` and the repr would be the one to add `proxied_list(...)`, but whatever

sharedbuffers/mapped_struct.py
```diff
@@ -320,6 +320,342 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
...
+
+if not cython.compiled:
+    setattr(proxied_list, '__eq__', getattr(proxied_list, '_eq'))
+    setattr(proxied_list, '__ne__', getattr(proxied_list, '_ne'))
+
+@cython.cclass
+class proxied_tuple(proxied_list):
+
+    def __str__(self):
+        return "proxied_tuple([%s])" % ",".join([str(x) for x in self])
```
klaussfreire: I'm pretty sure you need to implement more stuff for it to be a tuple-like thing.

You need a `__hash__` method for instance.

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
gefarion: I think that I have to use __dealloc__ if I use __cinit__ to initialize or allocate memory for the python object. But because I am using __init__, I can use __del__ to release the buffer.

sharedbuffers/mapped_struct.py
```diff
@@ -324,13 +331,17 @@ class mapped_dict(dict):
...
+        values = [obj[k] for k in keys]
+        return mapped_list.pack_into([keys, values], buf, offs, idmap, implicit_offs)
 
     @classmethod
     def unpack_from(cls, buf, offs, idmap = None):
         key, values = mapped_list.unpack_from(buf, offs, idmap)
         return cls(zip(key, values))
 
+@cython.ccall
+@cython.returns(cython.char)
```
klaussfreire: Why not int?

Also, this masks errors. Better add the declaration to the pxd and use the except syntax to specify an error value, so that it can throw exceptions.

sharedbuffers/mapped_struct.py
```diff
@@ -631,20 +638,20 @@ def __iter__(self):
             yield self[i]
 
     def __reversed__(self):
-        for i in reversed(xrange(len(self))):
+        for i in xrange(len(self) - 1, -1, -1):
```
klaussfreire: Since len() in Cython probably returns an unsigned type, better check that it isn't 0 before doing this.

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
gefarion: Why not char if the returns values are -1, 0 and 1?

klaussfreire: In my experience, you don't gain anything from using a narrower return type. In fact, the ABI is more cumbersome and slower for narrower types, `int` is C's native type and it tends to outperform narrower types. So unless there's a type congruence reason to use char (say, you want to match the type of some other API) or space reasons to narrow it down (say, it's going into an array or struct), you're better off using int.

sharedbuffers/mapped_struct.py
```diff
@@ -660,6 +668,23 @@ def __str__(self):
...
+        _hash = object,
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(proxied_tuple, self).__init__(*args, **kwargs)
+        self._hash = None
+
+    def __hash__(self):
+        if self._hash is None:
+            self._hash = hash(tuple(iter(self)))
```
``` python
mult = 1000003
x = 0x345678
len_ = len(self)
for ob in self:
    len -= 1
    y = hash(ob)
    x = (x ^ y) * mult;
    mult += 82520 + len + len;

x += 97531
if x == -1:
     x = -2
```

sharedbuffers/mapped_struct.py
```diff
@@ -665,20 +665,39 @@ def __str__(self):
...
-        _hash = object,
+        _hash = cython.long,
     )
 
     def __init__(self, *args, **kwargs):
         super(proxied_tuple, self).__init__(*args, **kwargs)
-        self._hash = None
+        self._hash = -1
 
+    @cython.locals(mult = cython.long, x = cython.long, y = cython.long, len_ = cython.size_t, i = cython.size_t)
```
klaussfreire: `cython.Py_ssize_t`

klaussfreire (changes requested):

klaussfreire (approved):

sharedbuffers/mapped_struct.py
```diff
@@ -321,6 +328,388 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
...
+        raise NotImplementedError
+
+#@cython.ccall
+#@cython.returns(int)
+def proxied_list_cmp(a, b):
+
+    alen = len(a)
+    blen = len(b)
+
+    if alen < blen:
```
tzulberti-jampp: You might be changing the Python implementation for list comparison:
```
>>> [4,5] < [1,2,3]
False
```

tzulberti-jampp (approved):

PR #42, opened 2018-09-26
Treat long as int, and produce more informative errors

klaussfreire:


sharedbuffers/mapped_struct.py
```diff
@@ -1174,6 +1175,7 @@ class ObjectBufferProxyProperty(GenericBufferProxyProperty):
     mapped_decimal : DecimalBufferProxyProperty,
 
     int : LongBufferProxyProperty,
+    long : LongBufferProxyProperty,
```
llogiudice, line 1178: I'm confused. Is this a Cython long (i.e: an int64), or is this a Python long ? If it's the latter, shouldn't there be support for arbitrarily long numbers ?

klaussfreire: It's a python long.

We should, but this isn't implementing that.

It just happens that we routinely get longs that fit into an int. So until we start handling them as true longs, we can handle most cases like this. If the value doesn't fit, it will raise an error anyway.

llogiudice (approved):

sharedbuffers/mapped_struct.py
klaussfreire: Hold on, though, I'm adding more error message improvements shortly.

gefarion (approved):

PR #47, opened 2018-10-01
Fix failing tests running on python mode

gefarion:


sharedbuffers/mapped_struct.py
```diff
@@ -422,7 +422,6 @@ def unpack_from(cls, buf, offs, idmap = None):
 # @cython.ccall
 # @cython.returns(cython.bint)
 @cython.locals(_cmp = int)
-@cython.optimize.use_switch(True)
```
tzulberti-jampp, line 425: Why did you remove this?

gefarion: It is true by default.

tzulberti-jampp (approved):

klaussfreire (approved):

PR #48, opened 2018-10-03
Add support for proxied dicts

llogiudice:
Ticket: https://jampphq.atlassian.net/browse/CORE-1854

### DIFF
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -364,8 +364,12 @@ def _stable_hash(key):
     elif isinstance(key, int):
         hval = key
     elif isinstance(key, float):
-        mant, expo = math.frexp(key)
-        hval = _mix_hash(expo, int(mant * 0xffffffffffff))
+        trunc_key = int(key)
+        if trunc_key == key:
+            hval = trunc_key
+        else:
+            mant, expo = math.frexp(key)
+            hval = _mix_hash(expo, int(mant * 0xffffffffffff))
     elif isinstance(key, (tuple, proxied_tuple, frozenset)):
         hval = xxhash.xxh64(str(type(key))).intdigest()
         for value in key:
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -340,7 +340,7 @@ class mapped_dict(dict):
     @classmethod
     def unpack_from(cls, buf, offs, idmap = None):
         proxy = proxied_dict.unpack_from(buf, offs, idmap)
-        return {k: v for k, v in proxy.iteritems()}
+        return proxy.copy()
 
 
 def _upsize(x):
@@ -350,17 +350,25 @@ def _upsize(x):
     return x + 1
 
 
+@cython.ccall
+@cython.locals(code=cython.ulong, nbits=cython.ulong)
 def _hash_rotl(code, nbits):
     return (code << nbits) | (code >> (32 - nbits))
 
 
+@cython.ccall
+@cython.locals(code1=cython.ulong, code2=cython.ulong)
 def _mix_hash(code1, code2):
     return _hash_rotl(code1, 5) ^ code2
 
 
+_TYPE_SEEDS = {tuple: 4110121799, frozenset: 3377403095}
+
+@cython.ccall
+@cython.locals(hval=cython.ulong)
 def _stable_hash(key):
     if isinstance(key, basestring):
-        hval = xxhash.xxh64(key).intdigest()
+        hval = xxhash.xxh64(safe_utf8(key)).intdigest()
     elif isinstance(key, int):
         hval = key
     elif isinstance(key, float):
@@ -370,16 +378,14 @@ def _stable_hash(key):
         else:
             mant, expo = math.frexp(key)
             hval = _mix_hash(expo, int(mant * 0xffffffffffff))
-    elif isinstance(key, (tuple, proxied_tuple, frozenset)):
-        hval = xxhash.xxh64(str(type(key))).intdigest()
+    elif type(key) in _TYPE_SEEDS:
+        hval = _TYPE_SEEDS[type(key)]
         for value in key:
             hval = _mix_hash(hval, _stable_hash(value))
     else:
-        raise TypeError("don't know how to hash key of type %s" % type(key))
+        raise TypeError("unhashable type: %s" % type(key).__name__)
 
-    # Truncating the hash value to 48 bits should have no practical impact
-    # on keys' destribution, and it helps us catch subtle overflow bugs
-    return hval & 0xffffffffffff
+    return hval
 
 
 def _enum_keys(obj):
@@ -420,6 +426,8 @@ class BufferIO(object):
 @cython.cclass
 class proxied_dict(object):
 
+    HEADER_PACKER = struct.Struct('=Q')   # Offset into value list.
+
     cython.declare(objmapper=object, vlist=proxied_list)
 
     def __init__(self, objmapper, vlist):
@@ -428,18 +436,20 @@ class proxied_dict(object):
 
     @classmethod
     def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0):
+        packer = cls.HEADER_PACKER
         ipos = offs
-        offs += 8   # Reserve space for value-list offset.
+        offs += packer.size
         iobuf = BufferIO(buf, offs)
         mapper_size = ObjectIdMapper.build(_enum_keys(obj), iobuf, map_file=False)
         offs += mapper_size
-        struct.pack_into("=Q", buf, ipos, mapper_size)
+        packer.pack_into(buf, ipos, mapper_size)
         return offs + proxied_list.pack_into(obj.values(), buf, offs, idmap, implicit_offs)
 
     @classmethod
     def unpack_from(cls, buf, offs, idmap = None):
-        mapper_size, = struct.unpack_from("=Q", buf, offs)
-        offs += 8
+        packer = cls.HEADER_PACKER
+        mapper_size, = packer.unpack_from(buf, offs)
+        offs += packer.size
         objmapper = ObjectIdMapper.map_buffer(buf, offs)
         offs += mapper_size
         vlist = proxied_list.unpack_from(buf, offs, idmap)
@@ -456,7 +466,7 @@ class proxied_dict(object):
         return self.vlist[idx]
 
     def __contains__(self, key):
-        return self.objmapper.get(key) is not None
+        return key in self.objmapper
 
     def has_key(self, key):
         return key in self
@@ -472,19 +482,31 @@ class proxied_dict(object):
         return iter(self.objmapper)
 
     def keys(self):
-        return list(self.iterkeys())
+        return self.objmapper.keys()
 
     def itervalues(self):
         return iter(self.vlist)
 
     def values(self):
-        return list(self.itervalues())
+        return self.vlist
 
     def __iter__(self):
         return self.iterkeys()
 
     def __len__(self):
         return len(self.vlist)
+
+    def copy(self):
+        return dict(self.iteritems())
+
+    def viewkeys(self):
+        return self.keys()
+
+    def viewvalues(self):
+        return self.values()
+
+    def viewitems(self):
+        return self.items()
 
     def _is_eq(self, other):
         if len(self) != len(other):
@@ -4126,7 +4148,7 @@ class ObjectIdMapper(_CZipMapBase):
         return mapped_object.unpack_from(buf, index)
 
     @cython.locals(i = cython.ulonglong, indexbuf = 'Py_buffer', pybuf = 'Py_buffer')
-    def iterkeys(self, make_sequential = True):
+    def iterkeys(self, make_sequential = False):
         buf = self._buf
         dtype = self.dtype
         if make_sequential:
@@ -4192,7 +4214,7 @@ class ObjectIdMapper(_CZipMapBase):
 
     @cython.locals(i = cython.ulonglong, indexbuf = 'Py_buffer', pybuf = 'Py_buffer',
         stride0 = cython.size_t, stride1 = cython.size_t, pindex = cython.p_char)
-    def iteritems(self, make_sequential = True):
+    def iteritems(self, make_sequential = False):
         buf = self._buf
         dtype = self.dtype
         if make_sequential:
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -494,12 +494,19 @@ class proxied_dict(object):
 
         return True
 
-    def __richcmp__(self, other, op):
-        if op != Py_EQ and op != Py_NE:
-            return False   # No comparison possible
+    if cython.compiled:
+        def __richcmp__(self, other, op):
+            if op != Py_EQ and op != Py_NE:
+                return False   # No comparison possible
 
-        rv = self._is_eq(other)
-        return rv if op == Py_EQ else not rv
+            rv = self._is_eq(other)
+            return rv if op == Py_EQ else not rv
+    else:
+        def __eq__(self, other):
+            return self._is_eq(other)
+
+        def __ne__(self, other):
+            return not self._is_eq(other)
 
     def __str__(self):
         return "proxied_dict({%s})" % ", ".join(["%s: %s" % (k, v) for k, v in self.iteritems()])
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -14,6 +14,8 @@ import xxhash
 import itertools
 import time
 import zipfile
+import math
+import sys
 from datetime import timedelta, datetime, date
 from decimal import Decimal
 import numpy as np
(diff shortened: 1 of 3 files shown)


tests/mapped_struct.py
```diff
@@ -143,6 +143,7 @@ class ContainerStruct(TestStruct):
 class DictStruct(TestStruct):
     __slot_types__ = {
         'd' : dict,
+        'D' : mapped_struct.proxied_dict
```
gefarion, line 146: You have to change `testDictStruct(self)` to use the new dict.

tests/mapped_struct.py
```diff
@@ -1619,6 +1620,19 @@ class MappedDictPackingTest(unittest.TestCase, CollectionPackingTestHelpers, Dic
...
+class ProxiedDictPackingTest(unittest.TestCase, CollectionPackingTestHelpers, DictPackingCommonTest):
+    PACKING_CLASS = mapped_struct.proxied_dict
+    COLLECTION_CLASS = dict
+
+    TEST_DICTS = [
+        {},
+        {'a': 'a2', 'b': 'b2', 'c': 'c2'},
+        {1: 10, 2: 20, 3: 30},
+        {'a': frozenset(), 'b': (1, 2), 'c': 1.0, 'd': [1, 2], 'e': dict(a=1) },
+        {1: 42, 'a1_@!': 69, 3.5: 'uhhhh'}
```
gefarion: Add some case using tuples and frozensets as keys.

sharedbuffers/mapped_struct.py
```diff
@@ -342,6 +342,194 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+
+def _upsize(x):
+    "Round up X to the next power of 2"
+    for p2 in (1, 2, 4, 8, 16, 32, 64):
+        x |= x >> p2
+    return x + 1
+
+def _add_kv(indexes, kvpairs, key, val):
+    l2 = len(indexes) / 2
+    hval = hash(key)
```
klaussfreire: You cannot use python's hash. It can be randomized, it's only guaranteed to be stable within the process, but not across process boundaries.

You need to build your own stable hash function.

sharedbuffers/mapped_struct.py
```diff
@@ -342,6 +342,194 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+        return offs + mapped_tuple.pack_into(kvpairs, buf, offs, idmap, implicit_offs)
+
+    @classmethod
+    def unpack_from(cls, buf, offs, idmap = None):
+        mask, nelems = struct.unpack_from("=QQ", buf, offs)
+        offs += 16
+        idx_size = (mask + 1) * 16
+        index_table = array.array('l')
+        index_table.fromstring(buffer(buf[offs:offs + idx_size]))
+        index_table = tuple(index_table)
```
klaussfreire: This is not a proxied dict. In here you're materializing the whole dict in memory. I very much prefer mapped_dict's implementation in that case, so thanks but no thanks.

The whole point of proxies is to avoid copying the whole data structure into process private memory (ie: copying). The `unpack_from` operation must be O(1), it must construct a class that can then look up the required item in less than O(N).

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
llogiudice: Isn't this only true as of python 3.3 ? https://docs.python.org/3/reference/datamodel.html#object.__hash__

I can make my own stable hash, but doesn't that mean restricting the key types beforehand ?

sharedbuffers/mapped_struct.py
llogiudice: Would it be enough to change the index table into a numpy array and then use that proxy ?

sharedbuffers/mapped_struct.py
klaussfreire: If that were the case, it still applies, since we want to eventually be compatible with Python 3. Also, shared buffers must have a portable binary format, and the hash function used affects the usability of the format, so you want a portable hash function.

And no. You can enable randomized hashes in Python 2 as well.

sharedbuffers/mapped_struct.py
klaussfreire: No. The current lookup functions are broken when rehashing nontrivial hash maps with collisions (the probe sequence isn't preserved crossing the 2^n boundary). You need a correct implementation, and hash tables are tricky and very unfriendly for disk access.

Sharedbuffers are mapped from disk and they can be bigger than RAM, so we have an algorithm already that is optimized for disk friendliness, compactness and is less than O(N), which is the one the various IdMappers use.

Use the id mappers. If not the classes at least the algorithms and helpers.

klaussfreire: Though I could have misread. Reading again. Still, I'd rather reuse the id mappers.

gefarion (changes requested):

sharedbuffers/mapped_struct.py
```diff
@@ -342,6 +342,194 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+        index_table=tuple,
+        kvpairs=proxied_list,
+        mask=int,
+        nelems=int)
+
+    def __init__(self, mask, nelems, index_table, kvpairs):
+        self.mask = mask
+        self.nelems = nelems
+        self.index_table = index_table
+        self.kvpairs = kvpairs
```
klaussfreire: For this to be a proxy, you can't have all the items as python objects. You need to only store a pointer to the position in the buffer where the data is, and then implement all getters by parsing the on-disk structure.

sharedbuffers/mapped_struct.py
```diff
@@ -342,6 +342,194 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+        The lookup algorithm computes a position by doing hash(key) modulo len(index_vector).
+        Since the index vector is always rounded to a power of 2, this is a
+        very fast operation.
+
+        In addition, before comparing keys, we do a quick check against the
+        saved hash value. Because the same restrictions as with Python apply,
+        said operation is stable, and must match if the keys compare equal.
+
+        Collisions are resolved with a combination of open addressing and
+        linear probing.
```
klaussfreire: You have 2 implementations of this. You really should have only one.

I'd propose you use `_search_hkey` (one of the variants) for this. It has better worst-case performance than that of a hash table (which is O(N)), and even lower than O(log N) most of the time, with O(1) disk I/Os, which means it's a good fit for memory-mapped files.

The on-disk format is also more compact.

sharedbuffers/mapped_struct.py
```diff
@@ -342,6 +342,194 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+        In other words, there are 2 tables: an index vector and a key/value
+        array. The index vector contains the positions in the second array,
+        and their hash values, to speed up comparisons. Unused buckets have
+        a position of -1. The key/value array is packed into a proxied list.
+        """
+
+        xlen = len(obj)
+        size = _upsize(xlen + xlen / 2)   # Aim for a load factor of ~50%
+        indexes = [-1, 0] * size
+        kvpairs = []
```
klaussfreire: There's no need to keep a list of kvpairs.

You know the size of the index beforehand (in any case, with the hash table or the sorted table `_search_hkey` uses). You can leave enough room in the destination buffer for the index and pack the key-value pairs as you add the items to the index. Then you sort the index in-place and you're done.

sharedbuffers/mapped_struct.py
```diff
@@ -342,6 +342,194 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+        nelem = 0
+
+        for key, val in obj.iteritems():
+            _add_kv(indexes, kvpairs, key, val)
+            nelem += 1
+
+        len_m1 = len(indexes) / 2 - 1
+        struct.pack_into("=QQ", buf, offs, len_m1, nelem)
+        offs += 16
+        abuf = buffer(array.array('l', indexes))
```
klaussfreire: This is creating several copies of the index. It's *very* memory-intensive. You need to be mindful of the resources you use.

klaussfreire (changes requested):
Access to indexes and headers should be cython-optimized. Cython can use pointer arithmetic and dereference, along with the PyBuffer interface, to avoid using struct and copying things around. Sharedbuffers is all about low-overhead, so you'll have to do a cython-optimized version and a pure-python version for these cases where doing pointer arithmetic can save you thousands of operations.

struct.unpack: thousands of cycles in object creation, dynamic method calls, attribute access, parsing the format struct, etc

(<int*>buf.buffer)[0] = 1 cycle + memory latency.

sharedbuffers/mapped_struct.py
llogiudice: I'm delegating this to proxied_list. Doesn't it handle that ?

tests/mapped_struct.py
llogiudice: what do you mean ? DictStruct already contains a reference to proxied_dict in __slot_types__

sharedbuffers/mapped_struct.py
llogiudice: Doesn't this limit us to only numeric keys or values ?

klaussfreire: No. Hashes are always numeric. Values are item offsets.

sharedbuffers/mapped_struct.py
klaussfreire: You built a tuple with the index.

The kvpairs proxied list contains tuples (regular tuples) inside it, when you access an item you'll be building that tuple.

You've got lots of holes in that approach.

llogiudice: Huh ? The kvpairs proxied_list is flat. It doesn't contain tuples unless the dict itself has tuples as keys or values.

klaussfreire: Ah, yes, you're using extend. Still moot, you've copied the index into a tuple.

Just assume you've got a several-GB dict you've mmaped from a file here, and you still want instant unpack and instant access to an item.

Given this schema:

``` python
class Node:
    __slot_types__ = {
        'branches' : proxied_dict
    }
    __schema__ = Schema.from_typed_slots(__slot_types__)
```

Build a really huge proxied_dict and compare the performance of doing:

``` python
with open('somefile', 'rb') as f:
    f.seek(0, os.SEEK_END)
    size = f.tell()
    buf = mmap.mmap(f.fileno(), size, access = mmap.ACCESS_READ)
    node = node.__schema__.unpack_from(buf)
    node.branches[key]
```

It should be really fast, independent of file size, and use almost no memory.

We'll be using this proxied_dict in the BDT nodes, so item access time should be comparable with python dict access.

I'm pretty sure this is not the case in this implementation.

sharedbuffers/mapped_struct.py
klaussfreire: Se how StringIdMapper does its thing. You'd do something similar:

Pack a matrix of:
```
[ hash0  key_off   val_off ]
...
[ hashN  key_off   val_off ]
```

Pack each key/val and populate the matrix, then sort the matrix by hash. You can do all that with numpy efficiently.

Then on reading `search_hkey` will help you get the first matching hash, then you do a sequential search to find the matching key (should be a small run since they'd all be hash collisions).

All of that is in StringIdMapper. You can factor out the common bits and reuse them (careful not to hurt the mapper's performance in the process, use Cython annotations properly).

llogiudice: My problem comes with the sequential search part, because search_hkey returns the found value (or the default, if not present), instead of, say, an index (which it can't return). In that case, the lookup is downgraded to a linear search.

klaussfreire: Did you read `_search_hkey`?

klaussfreire:
Ping?

llogiudice:
> Ping?

I have the code working with 2 proxied lists (one for indices, the other for key/values). It's using linear probing, and it doesn't appear to be memory intensive, or have bad distributions with the stable hash function I've written.

Working on adapting StringIdMapper, however, has proven quite difficult. The original assumptions made about the stored values are hard to abstract. While I've gotten lookups to work with generic objects, iteration is a different story. The problem is that I can't know in advance the types of the keys and values, and have to store that information somewhere. So I'm looking into that right now, although I still think 2 proxied lists should work just as fine.

klaussfreire:
> > Ping?
> 
> I have the code working with 2 proxied lists (one for indices, the other for key/values). It's using linear probing, and it doesn't appear to be memory intensive, or have bad distributions with the stable hash function I've written.
> 
> Working on adapting StringIdMapper, however, has proven quite difficult. The original assumptions made about the stored values are hard to abstract. While I've gotten lookups to work with generic objects, iteration is a different story. The problem is that I can't know in advance the types of the keys and values, and have to store that information somewhere. So I'm looking into that right now, although I still think 2 proxied lists should work just as fine.

I would guess that for an ObjectIdMapper, the type of the keys would always be `mapped_object`. That class already handles the task of storing and reading type information.

sharedbuffers/mapped_struct.py
```diff
@@ -342,6 +344,167 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+            return False
+
+        for key, val in self.iteritems():
+            if key not in other or other[key] != val:
+                return False
+
+        return True
+
+    def __richcmp__(self, other, op):
+        if op != Py_EQ and op != Py_NE:
```
gefarion: These constants are not defined in Python.

sharedbuffers/mapped_struct.py
```diff
@@ -342,6 +344,167 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+            return False   # No comparison possible
```
gefarion: I think that if you can't compare the objects, you have to raise `NotImplementedError`

tests/mapped_struct.py
gefarion: Check line https://github.com/jampp/sharedbuffers/blob/82a2f0d3f08208462fbabccbecb480f99b3b31c5/tests/mapped_struct.py#L226

sharedbuffers/mapped_struct.py
llogiudice: Python 2 always returns False for operations without meaning (e.g: {} < {}), whereas Python3 raises a TypeError. I'm not sure what's the best option, but I think that since we're aiming to replicate the original dict behavior, going with Python2's behavior is saner (for now).

gefarion (approved):

sharedbuffers/mapped_struct.py
klaussfreire: Why don't you just implement `__eq__` and `__ne__` then?

sharedbuffers/mapped_struct.py
```diff
@@ -494,12 +494,19 @@ def _is_eq(self, other):
 
         return True
 
-    def __richcmp__(self, other, op):
-        if op != Py_EQ and op != Py_NE:
-            return False   # No comparison possible
+    if cython.compiled:
```
klaussfreire: This conditional method declaration will confuse Cython. You have to always define `__richcmp__`, do the conditional inside.

sharedbuffers/mapped_struct.py
llogiudice: > Why don't you just implement `__eq__` and `__ne__` then?

Cython barfs at you if you use those instead of __richcmp__

sharedbuffers/mapped_struct.py
```diff
@@ -335,14 +335,12 @@ class mapped_dict(dict):
...
-        values = [obj[k] for k in keys]
-        return mapped_list.pack_into([keys, values], buf, offs, idmap, implicit_offs)
+        return proxied_dict.pack_into(obj, buf, offs, idmap, implicit_offs)
 
     @classmethod
     def unpack_from(cls, buf, offs, idmap = None):
-        key, values = mapped_list.unpack_from(buf, offs, idmap)
-        return cls(zip(key, values))
+        proxy = proxied_dict.unpack_from(buf, offs, idmap)
+        return {k: v for k, v in proxy.iteritems()}
```
``` python
return dict(proxy.iteritems())
```

Or perhaps simply

``` python
return proxy.copy()
```

Since you do have to implement `copy` to fully behave as a dict.

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
klaussfreire: Which version of Cython are you using?

llogiudice: > Which version of Cython are you using?

0.26

sharedbuffers/mapped_struct.py
llogiudice: Looking at the annotated html, there doesn't seem to be anything wrong with it.

sharedbuffers/mapped_struct.py
```diff
@@ -3869,6 +4049,447 @@ def merge(cls, parts, destfile = None, tempdir = None,
...
+                # Not *quite* as correct as computing the actual length in
+                # bytes, but since this is only an upper bound, we should be fine.
+                klen = sys.getsizeof(k)
+                if curpos > dtypemax:
+                    raise ValueError("Cannot represent offset with requested precision")
+                part.append((_stable_hash(k), curpos, i))
+                if klen + 16 > valbuflen:
+                    valbuflen = (klen + 16) * 2
+                    valbuf = bytearray(valbuflen)
+                endpos = pack_into(k, valbuf, 0)
```
klaussfreire: Why do you use a temp bytearray instead of simply packing into the destination buffer?

sharedbuffers/mapped_struct.py
```diff
@@ -333,14 +335,185 @@ class mapped_dict(dict):
...
+    return (code << nbits) | (code >> (32 - nbits))
+
+
+def _mix_hash(code1, code2):
+    return _hash_rotl(code1, 5) ^ code2
+
+
+def _stable_hash(key):
+    if isinstance(key, basestring):
+        hval = xxhash.xxh64(key).intdigest()
```
klaussfreire: I think you have to treat unicode and bytes differently

sharedbuffers/mapped_struct.py
```diff
@@ -333,14 +335,185 @@ class mapped_dict(dict):
...
+
+def _hash_rotl(code, nbits):
+    return (code << nbits) | (code >> (32 - nbits))
+
+
+def _mix_hash(code1, code2):
+    return _hash_rotl(code1, 5) ^ code2
+
+
+def _stable_hash(key):
```
klaussfreire, line 367: These functions should be highly optimized. Meaning, at least, fast cython calling.

sharedbuffers/mapped_struct.py
```diff
@@ -333,14 +335,185 @@ class mapped_dict(dict):
...
+    if isinstance(key, basestring):
+        hval = xxhash.xxh64(key).intdigest()
+    elif isinstance(key, int):
+        hval = key
+    elif isinstance(key, float):
+        mant, expo = math.frexp(key)
+        hval = _mix_hash(expo, int(mant * 0xffffffffffff))
```
klaussfreire: I'm not convinced this works. A tricky part of hashing a float is that different representations of the same number must hash to the same value. Did you check what CPython does for inspiration?

sharedbuffers/mapped_struct.py
```diff
@@ -333,14 +335,185 @@ class mapped_dict(dict):
...
+    elif isinstance(key, (tuple, proxied_tuple, frozenset)):
+        hval = xxhash.xxh64(str(type(key))).intdigest()
```
klaussfreire: This is most certainly not a good idea. `str(type(key))` could be unstable.

Check `proxied_tuple`, you have an algorithm there I believe you can use. `proxied_tuple` uses python's hash as a base, but you can use the same algorithm replacing `hash` with `stable_hash`.

sharedbuffers/mapped_struct.py
```diff
@@ -333,14 +335,185 @@ class mapped_dict(dict):
...
+        for value in key:
+            hval = _mix_hash(hval, _stable_hash(value))
+    else:
+        raise TypeError("don't know how to hash key of type %s" % type(key))
```
klaussfreire: More than `don't know how to hash`, the usual message is "unhashable type"

sharedbuffers/mapped_struct.py
```diff
@@ -333,14 +335,185 @@ class mapped_dict(dict):
...
+
+    # Truncating the hash value to 48 bits should have no practical impact
```
klaussfreire: Don't, use overflow-safe math

sharedbuffers/mapped_struct.py
```diff
@@ -333,14 +335,185 @@ class mapped_dict(dict):
...
+        offs += 8   # Reserve space for value-list offset.
+        iobuf = BufferIO(buf, offs)
+        mapper_size = ObjectIdMapper.build(_enum_keys(obj), iobuf, map_file=False)
+        offs += mapper_size
+        struct.pack_into("=Q", buf, ipos, mapper_size)
+        return offs + proxied_list.pack_into(obj.values(), buf, offs, idmap, implicit_offs)
+
+    @classmethod
+    def unpack_from(cls, buf, offs, idmap = None):
+        mapper_size, = struct.unpack_from("=Q", buf, offs)
```
klaussfreire: Do as all other classes do, put the header Struct as a class attribute and reuse it

sharedbuffers/mapped_struct.py
```diff
@@ -333,14 +335,185 @@ class mapped_dict(dict):
...
+        offs += 8
```
klaussfreire: Use the struct's `.size`

sharedbuffers/mapped_struct.py
```diff
@@ -333,14 +335,185 @@ class mapped_dict(dict):
...
+        return self.vlist[idx] if idx is not None else default_val
+
+    def __getitem__(self, key):
+        idx = self.objmapper.get(key)
+        if idx is None:
+            raise KeyError(key)
+        return self.vlist[idx]
+
+    def __contains__(self, key):
+        return self.objmapper.get(key) is not None
```
klaussfreire: I think you can do `key in self.objmapper`

sharedbuffers/mapped_struct.py
```diff
@@ -333,14 +335,185 @@ class mapped_dict(dict):
...
+            yield key, self.vlist[idx]
+
+    def items(self):
+        return list(self.iteritems())
+
+    def iterkeys(self):
+        return iter(self.objmapper)
+
+    def keys(self):
+        return list(self.iterkeys())
```
klaussfreire: You can just return `self.objmapper.keys()`

Also, check that you override the default `make_sequential=True` with False, I think it can be surprising if iterating a dict has to sort big arrays.

sharedbuffers/mapped_struct.py
```diff
@@ -333,14 +335,185 @@ class mapped_dict(dict):
...
+
+    def itervalues(self):
+        return iter(self.vlist)
+
+    def values(self):
+        return list(self.itervalues())
```
klaussfreire: `return self.vlist`

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
klaussfreire: 0.27 supports it, maybe we should consider requiring 0.27.0 (0.27.1 has bugs that make it not build)

Otherwise, you can leave __richcmp__ unconditional *and* implement eq and ne, using the import-time monkeypatching trick you committed earlier.

klaussfreire: BTW, what python does isn't to return False all the time. It compres the `id` of objects instead.

sharedbuffers/mapped_struct.py
llogiudice: I'm using another IdMapper as a template, and I guess the reasoning there is that using a temparray is easier than writing directly to a file.

sharedbuffers/mapped_struct.py
llogiudice: Python 2 hashes bytes, strings and unicode all the same. To me, it's the same idea as with floats and integers: They should hash to the same value as long as they are equivalent.

sharedbuffers/mapped_struct.py
llogiudice: They do pretty much the same thing, except they process 28 bits at a time instead of 64 (because of stupid C promotion rules). I don't see any special handling to ensure that different representations hash to the same value, other than relying on frexp normalizing its return values.

sharedbuffers/mapped_struct.py
klaussfreire: Sorry, yes, I thought build was building the idmapper into a buffer, but you're right, it's using a file.

sharedbuffers/mapped_struct.py
```diff
@@ -3869,6 +4049,447 @@ def merge(cls, parts, destfile = None, tempdir = None,
...
+        n = 0
+        part = []
+        while 1:
+            del part[:]
+            for k,i in islice(initializer, 1000):
+                # Add the index item
+                n += 1
+                # Not *quite* as correct as computing the actual length in
+                # bytes, but since this is only an upper bound, we should be fine.
+                klen = sys.getsizeof(k)
```
klaussfreire: Trouble is, you need the recursive size. This will probably fall short lots of times, and you're not retrying the way Schema.pack does.

You probably will have to implement the same retry mechanism schema.Pack uses, and you should also try to better estimate the size of the required buffer. You could have a configurable "default minimum buffer size" defined somewhere.

You can also reuse the bytearray for each object in the loop, you don't need a new byte array for each key. So you could just avoid trying to guess the right buffer size, and just start with a configurable buffer and expand the buffer when necessary (when pack_into raises an overflow exception, check Schema.pack).

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
klaussfreire: What python2 does here doesn't matter, it matters what xxhash does.

klaussfreire:
``` pycon
>>> xxhash.xxh64(u'holá').intdigest()
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe1' in position 3: ordinal not in range(128)
```

sharedbuffers/mapped_struct.py
klaussfreire: Oh... if frexp noramlizes, then yes, this should handle that.

sharedbuffers/mapped_struct.py
```diff
@@ -543,8 +549,11 @@ def _eq(self, other):
         def _ne(self, other):
             return not self._is_eq(other)
 
+    def __repr__(self):
+        return "proxied_dict(%s)" % self
+
     def __str__(self):
-        return "proxied_dict({%s})" % ", ".join(["%s: %s" % (k, v) for k, v in self.iteritems()])
+        return "{%s}" % ", ".join(["%s: %s" % (k, v) for k, v in self.iteritems()])
```
klaussfreire: You have to use `"%r: %r" for correct results

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
```diff
@@ -333,14 +335,231 @@ class mapped_dict(dict):
...
+
+_TUPLE_SEED = 4110121799
+_FSET_SEED  = 3377403095
+
+@cython.ccall
+@cython.locals(hval=cython.ulong)
+def _stable_hash(key):
+    if isinstance(key, basestring):
+        hval = xxhash.xxh64(safe_utf8(key)).intdigest()
+    elif isinstance(key, int):
```
klaussfreire: `(int, long)` for python 2's sake

sharedbuffers/mapped_struct.py
```diff
@@ -333,14 +335,231 @@ class mapped_dict(dict):
...
+        hval = key
+    elif isinstance(key, float):
+        trunc_key = int(key)
+        if trunc_key == key:
+            hval = trunc_key
+        else:
+            mant, expo = math.frexp(key)
```
klaussfreire, line 377: Can you make Cython use C's native frexp?

You'll probably have to add an `if cython.compiled` here, because C's frexp is radically different in calling convention. But it should be worth the trouble.

sharedbuffers/mapped_struct.py
```diff
@@ -333,14 +335,231 @@ class mapped_dict(dict):
...
+            hval = _mix_hash(hval, _stable_hash(value))
+    else:
+        raise TypeError("unhashable type: %s" % type(key).__name__)
+
+    return hval
+
+
+def _enum_keys(obj):
+    idx = 0
+    for key in obj.iterkeys():
```
``` python
@cython.locals(idx = int)
...
for idx, key in enumerate(obj):
```

sharedbuffers/mapped_struct.py
```diff
@@ -333,14 +335,231 @@ class mapped_dict(dict):
...
+        self.objmapper = objmapper
+        self.vlist = vlist
+
+    @classmethod
+    def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0):
+        packer = cls.HEADER_PACKER
+        ipos = offs
+        offs += packer.size
+        iobuf = BufferIO(buf, offs)
+        mapper_size = ObjectIdMapper.build(_enum_keys(obj), iobuf, map_file=False)
```
klaussfreire: Maybe you should implement the `map_file=False` option in all id mapper classes to have a consistent convention

And I'd rather call it `return_mapper=False`

sharedbuffers/mapped_struct.py
```diff
@@ -333,14 +335,231 @@ class mapped_dict(dict):
...
+        offs += mapper_size
```
klaussfreire: The build doesn't have to return the size, you could do `iobuf.tell()`

sharedbuffers/mapped_struct.py
klaussfreire: @gefarion experienced the issue, it's not in the annotated html because what happens is that it doesn't generate the required fields in the type descriptor. The issue is in the boilerplate code it generates to describe the type for Python. It basically won't invoke this method when doing comparisons.

You have to do it like:

```  python
class Whatever:
    def __richcmp__(self, other, op):
        ...

    if not cython.compiled:
        def _cmp(self, other):
            ...

        def _ne(self, other):
            ...

        def _eq(self, other):
            ...

if not cython.compiled:
    for orig, new in (('_cmp', '__cmp__'), ('_eq', '__eq__'), ('_ne', '__ne__')):
        setattr(proxied_dict, new, getattr(proxied_dict, orig))
```

So only what's used in pure python mode is conditional on `cython.compiled`.

Pure python will have the richcmp method, but it doesn't hurt. And you can alwas delattr if you need to.

sharedbuffers/mapped_struct.py
```diff
@@ -3869,6 +4056,454 @@ def merge(cls, parts, destfile = None, tempdir = None,
...
+        if offset:
+            self._buf = self._likebuf = buffer(buf, offset)
+        else:
+            self._buf = buf
+            self._likebuf = _likebuffer(buf)
+
+        # Parse header and map index
+        self.index_elements, self.index_offset = self._Header.unpack_from(self._buf, 0)
+
+        self.index = numpy.ndarray(buffer = self._buf,
```
klaussfreire: First line should have no arguments. I know, we've sinned before.

sharedbuffers/mapped_struct.py
```diff
@@ -3869,6 +4056,454 @@ def merge(cls, parts, destfile = None, tempdir = None,
...
+            dtype = self.dtype
+            struct_dt = numpy.dtype([
+                ('key_hash', dtype),
+                ('key_offset', dtype),
+                ('value', dtype),
+            ])
+            return self.index.view(struct_dt).reshape(self.index.shape[0]).searchsorted(
+                numpy.array([(hkey,0,0)],dtype=struct_dt))[0]
+
+    def _compare_keys(self, buf, offs, key):
```
klaussfreire, line 4338: @cython.ccall ?

sharedbuffers/mapped_struct.py
```diff
@@ -3869,6 +4056,454 @@ def merge(cls, parts, destfile = None, tempdir = None,
...
+        return self.index_elements
+
+    def __contains__(self, key):
+        return self.get(key) is not None
+
+    def preload(self):
+        # Just touch everything in sequential order
+        self.index.max()
+
+    def _unpack(self, buf, index):
```
klaussfreire, line 4157: @cython.ccall ?

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
llogiudice: I don't know about this. In additon to frexp having an output parameter, which I don't know how it translates to cython, there's the complication that we'd also have to deal with +/Inf and NaN.

Looking at the source, it appears that CPython uses a straightforward translation of C's native call, but at the same time handling the above problems:

https://github.com/python/cpython/blob/2.7/Modules/mathmodule.c#L1155

I can try to cythonize this, but I don't think it's really worth it. Besides, how many floats do we see as keys in hash tables anyway ?

sharedbuffers/mapped_struct.py
```diff
@@ -346,7 +346,7 @@ def unpack_from(cls, buf, offs, idmap = None):
 @cython.ccall
 @cython.locals(code=cython.ulong, nbits=cython.ulong)
```
klaussfreire: For 64-bit hashes you need `ulonglong` (for portability)

sharedbuffers/mapped_struct.py
```diff
@@ -387,10 +387,8 @@ def _stable_hash(key):
 
 
 def _enum_keys(obj):
```
klaussfreire, line 394: Declare idx as int to get an optimized enumerate

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
```diff
@@ -344,13 +344,13 @@ def unpack_from(cls, buf, offs, idmap = None):
...
 @cython.ccall
-@cython.locals(code=cython.ulong, nbits=cython.ulong)
+@cython.locals(code=cython.ulonglong, nbits=cython.ulonglong)
 def _hash_rotl(code, nbits):
     return (code << nbits) | (code >> (64 - nbits))
 
 
 @cython.ccall
-@cython.locals(code1=cython.ulong, code2=cython.ulong)
+@cython.locals(code1=cython.ulonglong, code2=cython.ulonglong)
```
klaussfreire, line 357: Should also `@cython.returns(cython.ulonglong)` right?

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
llogiudice: I thought cython was smart enough to figure that one out ...

Oh well, adding it won't hurt.

klaussfreire: It can probably infer it, but it will never change the function's signature and calling convention, which is what we're after.

klaussfreire: Also, unless you're unitt-testing these, you can use `cfunc` instead of `ccall` which is even lighter-weight. Since these are really tiny functions it probably helps.

You can also `cython.inline`.

klaussfreire (approved):

PR #49, opened 2018-10-03
Fix mapped_dict idmap handling

klaussfreire:
It leaves ephemeral objects (keys, values) in the idmap, causing
all kinds of weird mixups

### DIFF
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -281,6 +281,7 @@ class mapped_list(list):
         return mapped_tuple.pack_into(obj, buf, offs, idmap, implicit_offs)
 
     @classmethod
+    @cython.locals(rv = list)
     def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
                 dtype : array.array(dtype, []).itemsize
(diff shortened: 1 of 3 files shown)


sharedbuffers/mapped_struct.py
```diff
@@ -1027,24 +1038,28 @@ def unpack_from(cls, buf, offs, idmap = None, packer = struct.Struct('=q')):
         return rv
 
 class mapped_datetime(datetime):
+    PACKER = struct.Struct('=q')
+
     @classmethod
     @cython.locals(offs = cython.longlong, implicit_offs = cython.longlong, timestamp = cython.longlong)
-    def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0, packer = struct.Struct('=q')):
+    def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0):
```
tzulberti-jampp, line 1050: Why is this change only done for the datetime and not for other classes (like the date one)?

klaussfreire: Um...

tzulberti-jampp (approved):

gefarion (approved):

PR #52, opened 2018-10-16
Slices for proxied_list's

llogiudice:
Allow the use of slices and negative indices in proxied_list's (and by extension, proxied_tuple's).

### DIFF
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -841,12 +841,16 @@ class proxied_list(object):
                 xlen = (self.elem_end - self.elem_start - 1) / step + 1
             else:
                 xlen = (self.elem_start - self.elem_end - 1) / step + 1
+
+            if (self.elem_step < 0 and index > self.elem_start) or (
+                self.elem_step > 0 and index >= self.elem_end):
+                raise IndexError
             index = self.elem_start + index * self.elem_step
 
         if index < 0:
             index += xlen
 
-        if (index / step) >= max(xlen, objlen) or index < 0:
+        if index >= objlen or index < 0:
             raise IndexError
 
         if dcode == 't':
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -789,6 +789,12 @@ class proxied_list(object):
     def __init__(self, buf, offs, idmap = None, elem_start = 0, elem_end = 0, elem_step = 0):
         self.offs = offs
         self.buf = buf
+
+        if elem_step < 0:
+            elem_end = min(elem_end, elem_start)
+        elif elem_step > 0:
+            elem_start = min(elem_start, elem_end)
+
         self.elem_start = elem_start
         self.elem_end = elem_end
         self.elem_step = elem_step
@@ -888,6 +894,7 @@ class proxied_list(object):
 
         return res
 
+    @cython.ccall
     def _make_empty(self):
         return []
 
@@ -962,14 +969,16 @@ class proxied_list(object):
     def __delitem__(self, index):
         raise TypeError("Proxy objects are read-only")
 
+    @cython.locals(i=cython.longlong)
     def __iter__(self):
         for i in xrange(len(self)):
             yield self[i]
 
+    @cython.locals(l=cython.longlong)
     def __reversed__(self):
         l = len(self)
         if l == 0:
-            return self._make_empty()
+            return reversed(self._make_empty())
         for i in xrange(l - 1, -1, -1):
             yield self[i]
 
@@ -1002,6 +1011,7 @@ class proxied_tuple(proxied_list):
         super(proxied_tuple, self).__init__(*args, **kwargs)
         self._hash = -1
 
+    @cython.ccall
     def _make_empty(self):
         return ()
 
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -703,7 +703,9 @@ class proxied_list(object):
     cython.declare(
         buf = object,
         pybuf = 'Py_buffer',
-        offs = cython.ulonglong
+        offs = cython.ulonglong,
+        elem_start = cython.ulonglong,
+        elem_end = cython.longlong
     )
 
     if cython.compiled:
@@ -781,10 +783,13 @@ class proxied_list(object):
             else:
                 raise ValueError("Inconsistent data, unknown type code %r" % (dcode,))
 
-    @cython.locals(offs = cython.ulonglong, idmap = dict)
-    def __init__(self, buf, offs, idmap = None):
+    @cython.locals(offs = cython.ulonglong, idmap = dict, elem_start = cython.ulonglong,
+        elem_end = cython.longlong)
+    def __init__(self, buf, offs, idmap = None, elem_start = 0, elem_end = -1):
         self.offs = offs
         self.buf = buf
+        self.elem_start = elem_start
+        self.elem_end = elem_end
         self._init()
 
     @cython.ccall
@@ -812,13 +817,36 @@ class proxied_list(object):
         buf = _likerobuffer(buf)
         return cls(buf, offs, idmap)
 
-    @cython.locals(obj_offs = cython.ulonglong, dcode = cython.char, index = cython.ulonglong, pindex = "unsigned long *", dataoffs = cython.ulonglong)
-    def __getitem__(self, index):
+    def _getslice(self, start, end, step):
+        if step < 0:
+            while end >= start:
+                yield self[end + step]
+                end += step
+        else:
+            while start < end:
+                yield self[start]
+                start += step
+
+
+    @cython.locals(obj_offs = cython.ulonglong, dcode = cython.char, index = cython.longlong, pindex = "unsigned long *",
+        dataoffs = cython.ulonglong, xlen = cython.ulonglong)
+    def _getitem(self, index):
 
         dcode, objlen, itemsize, dataoffs, _struct = self._metadata()
+        xlen = objlen
 
-        if index >= objlen:
+        if self.elem_end >= 0:
+            xlen = self.elem_end - self.elem_start
+        else:
+            xlen = objlen
+
+        if index < 0:
+            index += xlen
+
+        if index >= xlen:
             raise IndexError
+
+        index += self.elem_start
 
         if dcode == 't':
             if cython.compiled:
@@ -864,6 +892,37 @@ class proxied_list(object):
 
         return res
 
+    def __getitem__(self, index):
+        if isinstance(index, slice):
+            start, end, step = index.start, index.stop, index.step
+            xlen = len(self)
+
+            if start is None:
+                start = 0
+            elif start < 0:
+                start = xlen + start
+
+            start += self.elem_start
+
+            if end is None:
+                end = xlen
+            elif end < 0:
+                end = xlen + end
+
+            if step is not None and step != 1:
+                return self._getslice(start, end, step)
+            elif start >= xlen:
+                return []
+            else:
+                return proxied_tuple(
+                    buf = self.buf,
+                    offs = self.offs,
+                    idmap = None,
+                    elem_start = start,
+                    elem_end = min(end, xlen))
+
+        return self._getitem(index)
+
     @cython.locals(op = cython.char)
     def  __richcmp__(self, other, op):
         return proxied_list_richcmp(self, other, op)
@@ -885,6 +944,8 @@ class proxied_list(object):
             return proxied_list_cmp(self, other) == 0
 
     def __len__(self):
+        if self.elem_end >= 0:
+            return self.elem_end - self.elem_start
         return self._metadata()[1]
 
     def __nonzero__(self):
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -831,7 +831,6 @@ class proxied_list(object):
 
         dcode, objlen, itemsize, dataoffs, _struct = self._metadata()
         xlen = objlen
-        step = 1
         orig_index = index
 
         if self.elem_step != 0:
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -704,8 +704,8 @@ class proxied_list(object):
         buf = object,
         pybuf = 'Py_buffer',
         offs = cython.ulonglong,
-        elem_start = cython.ulonglong,
-        elem_end = cython.ulonglong,
+        elem_start = cython.longlong,
+        elem_end = cython.longlong,
         elem_step = cython.longlong
     )
 
@@ -784,8 +784,8 @@ class proxied_list(object):
             else:
                 raise ValueError("Inconsistent data, unknown type code %r" % (dcode,))
 
-    @cython.locals(offs = cython.ulonglong, idmap = dict, elem_start = cython.ulonglong,
-        elem_end = cython.ulonglong, elem_step = cython.longlong)
+    @cython.locals(offs = cython.ulonglong, idmap = dict, elem_start = cython.longlong,
+        elem_end = cython.longlong, elem_step = cython.longlong)
     def __init__(self, buf, offs, idmap = None, elem_start = 0, elem_end = 0, elem_step = 0):
         self.offs = offs
         self.buf = buf
@@ -821,23 +821,26 @@ class proxied_list(object):
 
 
     @cython.locals(obj_offs = cython.ulonglong, dcode = cython.char, index = cython.longlong, pindex = "unsigned long *",
-        dataoffs = cython.ulonglong, xlen = cython.ulonglong)
+        dataoffs = cython.ulonglong, objlen = cython.longlong, xlen = cython.longlong, step = cython.longlong)
     def _getitem(self, index):
 
         dcode, objlen, itemsize, dataoffs, _struct = self._metadata()
         xlen = objlen
+        step = 1
 
         if self.elem_step != 0:
-            xlen = self.elem_end - self.elem_start
-            index *= self.elem_step
+            step = abs(self.elem_step)
+            if self.elem_step > 0:
+                xlen = (self.elem_end - self.elem_start - 1) / step + 1
+            else:
+                xlen = (self.elem_start - self.elem_end - 1) / step + 1
+            index = self.elem_start + index * self.elem_step
 
         if index < 0:
             index += xlen
 
-        if index >= xlen or index < 0 or index >= objlen:
+        if (index / step) >= max(xlen, objlen) or index < 0:
             raise IndexError
-
-        index += self.elem_start
 
         if dcode == 't':
             if cython.compiled:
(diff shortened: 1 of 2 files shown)


sharedbuffers/mapped_struct.py
```diff
@@ -781,10 +783,13 @@ def _metadata(self,
             else:
                 raise ValueError("Inconsistent data, unknown type code %r" % (dcode,))
 
-    @cython.locals(offs = cython.ulonglong, idmap = dict)
-    def __init__(self, buf, offs, idmap = None):
+    @cython.locals(offs = cython.ulonglong, idmap = dict, elem_start = cython.ulonglong,
+        elem_end = cython.longlong)
```
klaussfreire: elem_end should be ulonglong too, mixing them can cause undefined behavior.

Maybe initialize it to ~0 and just consider that it could be greater than the size of the list when operating with it.

Or use a triplet: `slice_start, slice_end, slice_step`, use `slice_step=0` as a flag to not slice at all (the whole sequence).

sharedbuffers/mapped_struct.py
```diff
@@ -812,14 +817,37 @@ def unpack_from(cls, buf, offs, idmap = None):
...
-    def __getitem__(self, index):
+    def _getslice(self, start, end, step):
+        if step < 0:
+            while end >= start:
+                yield self[end + step]
+                end += step
+        else:
+            while start < end:
+                yield self[start]
+                start += step
```
klaussfreire: No, a generator is not the idea. The idea is to returned a mapped_list that just points to a different slice of the buffer. Maybe you can play with strides to be able to support steps without much hardship.

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
```diff
@@ -835,10 +827,9 @@ def _getitem(self, index):
         dcode, objlen, itemsize, dataoffs, _struct = self._metadata()
         xlen = objlen
 
-        if self.elem_end >= 0:
+        if self.elem_step != 0:
             xlen = self.elem_end - self.elem_start
```
``` python
if self.elem_step != 0:
    if
    index = index * self.elem_step + self.elem_start
    if index < self.elem_start or index > self.elem_end:
        raise IndexError
```

sharedbuffers/mapped_struct.py
```diff
@@ -812,14 +819,26 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+
+        if self.elem_step != 0:
+            xlen = self.elem_end - self.elem_start
+            index *= self.elem_step
 
-        if index >= objlen:
+        if index < 0:
+            index += xlen
+
+        if index >= xlen or index < 0 or index >= objlen:
```
gefarion: I think that checking for `index >= objlen` is unnecessary.

klaussfreire: Yeah, @llogiudice said the same, but I'd rather have the check for safety. It's the only thing that clearly and inequivocally avoids a buffer overflow.

tests/mapped_struct.py
```diff
@@ -1500,6 +1500,15 @@ def testProxiedListRepr(self):
...
         self.assertEquals(self.pack([1, 2.0]), (1, 2.0))
 
+    def testProxiedListSlice(self):
+        orig = [0, '1', 2, '3', 4, '5']
+        obj = self.pack(orig)
+
+        self.assertEquals(obj[1:], orig[1:])
+        self.assertEquals(obj[:-1], orig[:-1])
+        self.assertEquals(obj[2:4], orig[2:4])
+        self.assertEquals(obj[1:-1:2], orig[1:-1:2])
```
gefarion, line 1511: - Add tests with huge start, end and step values.
- Add tests with negatives steps.
- Add test with only the step: `s[::-1]`, `s[::2]`.

gefarion (changes requested):

sharedbuffers/mapped_struct.py
```diff
@@ -812,12 +819,27 @@ def unpack_from(cls, buf, offs, idmap = None):
...
         dcode, objlen, itemsize, dataoffs, _struct = self._metadata()
+        xlen = objlen
+        step = 1
+
+        if self.elem_step != 0:
+            step = abs(self.elem_step)
+            if self.elem_step > 0:
+                xlen = (self.elem_end - self.elem_start - 1) / step + 1
+            else:
+                xlen = (self.elem_start - self.elem_end - 1) / step + 1
```
klaussfreire, line 843: This formula breaks if start==end

sharedbuffers/mapped_struct.py
```diff
@@ -864,6 +886,72 @@ def __getitem__(self, index):
...
+        val = min(val, nmax)
+        if step < 0 and val == nmax:
+            val -= 1
+        return val
+
+
+    def __getitem__(self, index):
+        if isinstance(index, slice):
+            start, end, step = index.start, index.stop, index.step
+            xlen = len(self)
```
klaussfreire, line 907: To make this more readable, why don't  you move this body to a `_getslice` method?

sharedbuffers/mapped_struct.py
```diff
@@ -864,6 +886,72 @@ def __getitem__(self, index):
...
+
+            if end is None:
+                end = xlen if step > 0 else -1
+            elif end < 0:
+                end = xlen + end
+                if end < 0:
+                    end = 0 if step > 0 else -1
+
+            end = self._adjust_min(end, xlen, step)
+            start = self._adjust_min(start, xlen, step)
```
klaussfreire: I think you can achieve all this with:

``` python
start, end, step = index.indices(len(self))
```

sharedbuffers/mapped_struct.py
```diff
@@ -864,6 +886,72 @@ def __getitem__(self, index):
...
+
+            if step < 0:
+                if end >= start:
+                    return []
+                r_len = (start - end - 1) / (-step) + 1
+                end = (r_len - 1) * step + start - 1
+            elif start >= end:
+                return ()
```
klaussfreire: Make up your mind. You either return an empty list or an empty tuple.

In fact, proxied_list should return an empty list and proxied_tuple should return an empty tuple.

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
llogiudice: If start == end, that will be caught by the __getitem__ method, and it will return an empty object.

sharedbuffers/mapped_struct.py
llogiudice: Yes, good idea.

sharedbuffers/mapped_struct.py
llogiudice: I wasn't aware of this method. Now I feel dumb.

sharedbuffers/mapped_struct.py
llogiudice: Good catch.

sharedbuffers/mapped_struct.py
klaussfreire: You cannot assume that. People can call the constructor explicitly, you don't expect Python to segfault when you pass a wrong argument, rather raise an error. So you cannot assume input will always be valid if the consecuence of invalid input is a segfault (which is what would happen if xlen is miscomputed)

llogiudice: Well, I disagree. If we went that route, we would have to check for buffer integrity everywhere for proxied objects, since they can be constructed that way too.

Regardless, I'll add the check since it's easy to do and not expensive at all.

klaussfreire: Well, yes. All classes do check for buffer integrity at risky places, AFAIK.

sharedbuffers/mapped_struct.py
```diff
@@ -829,6 +828,9 @@ def _getitem(self, index):
         step = 1
 
         if self.elem_step != 0:
+            if self.elem_end == self.elem_start:
+                # This shouldn't happen.
+                raise IndexError
```
klaussfreire: I think it should support it, it should behave as an empty list

sharedbuffers/mapped_struct.py
```diff
@@ -886,61 +888,27 @@ def _getitem(self, index):
...
 
-    @staticmethod
-    def _adjust_min(val, nmax, step):
-        val = min(val, nmax)
-        if step < 0 and val == nmax:
-            val -= 1
-        return val
-
+    def _make_empty(self):
+        return [] if type(self) is proxied_list else ()
```
klaussfreire: Why don't you just override it in proxied_list?

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
llogiudice: But if it's an empty list, any attempt at indexing is going to raise an IndexError exception, right ?

sharedbuffers/mapped_struct.py
llogiudice: Sure.

sharedbuffers/mapped_struct.py
klaussfreire: Yep. But other things won't, like iterating or taking len().

Which reminds me... you didn't implement `__nonzero__`

llogiudice: It's right under `__len__`

klaussfreire: Didn't see it :+1:

sharedbuffers/mapped_struct.py
```diff
@@ -885,7 +941,14 @@ def _eq(self, other):
             return proxied_list_cmp(self, other) == 0
 
     def __len__(self):
-        return self._metadata()[1]
+        if self.elem_step == 0:
+            return self._metadata()[1]
+        elif self.elem_start == self.elem_end:
```
tzulberti-jampp, line 950: shouldn't it return 0 if `self.elem_start <= self.elem_end`?

sharedbuffers/mapped_struct.py
```diff
@@ -943,6 +943,8 @@ def _eq(self, other):
     def __len__(self):
         if self.elem_step == 0:
             return self._metadata()[1]
+        elif self.elem_start == self.elem_end:
+            return 0
         elif self.elem_step < 0:
             return (self.elem_start - self.elem_end - 1) / (-self.elem_step) + 1
```
klaussfreire, line 953: What if `start < end`?

Or the converse for step > 0.

You should probably normalize that in the constructor:

``` python
if step < 0:
    end = min(start, end)
elif step > 0:
    start = min(start, end)
```

Then you don't need to worry about the in other places

sharedbuffers/mapped_struct.py
```diff
@@ -966,9 +968,10 @@ def __iter__(self):
 
     def __reversed__(self):
         l = len(self)
-        if l > 0:
-            for i in xrange(l - 1, -1, -1):
-                yield self[i]
+        if l == 0:
+            return self._make_empty()
```
klaussfreire: You can't return a value from a generator.

You can however stop iteration, by just `return`ing.

sharedbuffers/mapped_struct.py
```diff
@@ -966,9 +968,10 @@ def __iter__(self):
...
+        for i in xrange(l - 1, -1, -1):
```
klaussfreire: Declare `i` as `cython.longlong` to get an optimized iteration.

Same in other places where you do this.

sharedbuffers/mapped_struct.py
```diff
@@ -889,7 +889,7 @@ def _getitem(self, index):
         return res
 
     def _make_empty(self):
```
klaussfreire, line 902: Could make it ccall

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
llogiudice: Nah, those formulas work in most cases, _except_ when given malicious input, hence the need to manually handle them.

sharedbuffers/mapped_struct.py
llogiudice: That's a good idea, yes.

sharedbuffers/mapped_struct.py
llogiudice: Ah, right. I think I'll just emulate what happens with native lists and tuples.

sharedbuffers/mapped_struct.py
```diff
@@ -812,12 +825,29 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+        dataoffs = cython.ulonglong, objlen = cython.longlong, xlen = cython.longlong, step = cython.longlong)
+    def _getitem(self, index):
 
         dcode, objlen, itemsize, dataoffs, _struct = self._metadata()
+        xlen = objlen
+        step = 1
+
+        if self.elem_step != 0:
+            if self.elem_end == self.elem_start:
+                # This shouldn't happen.
```
klaussfreire: This comment is inaccurate. It can happen, it just happens that the right behavior here is to raise an IndexError.

tests/mapped_struct.py
klaussfreire: Ping @llogiudice

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
```diff
@@ -977,10 +977,9 @@ def __iter__(self):
     @cython.locals(l=cython.longlong)
     def __reversed__(self):
         l = len(self)
-        if l == 0:
-            return reversed(self._make_empty())
-        for i in xrange(l - 1, -1, -1):
-            yield self[i]
+        if l > 0:
+            for i in xrange(l - 1, -1, -1):
+                yield self[i]
```
tzulberti-jampp: If it is `empty` it is going to retrun None? Shouldn' it return a empty value?

llogiudice: Since this is a generator, the implicit returning of None is meant to stop iteration.

gefarion:
@llogiudice, I suggest adding a test case for every bug that you find.

sharedbuffers/mapped_struct.py
```diff
@@ -900,23 +900,16 @@ def _make_empty(self):
     def __getitem__(self, index):
         if isinstance(index, slice):
             xlen = len(self)
+            if index.step is not None and self.elem_step != 0:
+                index = slice(index.start, index.stop, index.step * self.elem_step)
+                xlen *= self.elem_step
```
klaussfreire: I don't see how this can work. I think this only works because your test doesn't use start/stop

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
llogiudice: Made some further adjustments for that, and an additional test. This is a lot more tricky than I thought.

sharedbuffers/mapped_struct.py
```diff
@@ -901,7 +901,9 @@ def __getitem__(self, index):
         if isinstance(index, slice):
             xlen = len(self)
             if index.step is not None and self.elem_step != 0:
-                index = slice(index.start, index.stop, index.step * self.elem_step)
+                start = index.start if index.start is None else index.start * self.elem_step
+                end = index.stop if index.stop is None else index.stop * self.elem_step
```
klaussfreire: Weird using `index.stop` when you know it's None.

Why not simply:

``` python
start = index.start
if start is not None:
    start *= self.elem_step

end = index.stop
if end is not None:
     end *= self.elem_step
```

It may be more code, but it's simpler to read

tests/mapped_struct.py
```diff
@@ -1515,6 +1515,7 @@ def testProxiedListSlice(self):
         self.assertEquals(obj[1::-1], orig[1::-1])
         self.assertEquals(obj[2:][:5], orig[2:][:5])
         self.assertEquals(obj[::2][::2], orig[::2][::2])
+        self.assertEquals(obj[1:10:2][2::2], orig[1:10:2][2::2])
```
klaussfreire: What of `obj[10:1:-2][2::2]`?

I think your code is too complex. You should simplify it (conceptually speaking).

The code should translate slice endpoints to array indexes, and then compound the steps. Something like:

``` python
start = self.elem_start + self.elem_step * index.start
end = self.elem_start + self.elem_step * index.end
step = self.elem_step * index.step
```

With all the caveats of negative indexes of course:

``` python
start = index.start
if start is not None:
    if start < 0:
        start += xlen
    start = self.elem_start + self.elem_step * index.start
    start = max(self.elem_start, min(self.elem_end, start))
```

Etc...

I don't know, I think that's cleaner and will be less prone to "special case bugs".

klaussfreire (changes requested):

tests/mapped_struct.py
llogiudice: I decided to use the original list size (i.e: the one encoded in the buffer) to compute the new indices. It made things simpler, I think.

sharedbuffers/mapped_struct.py
```diff
@@ -841,12 +841,16 @@ def _getitem(self, index):
                 xlen = (self.elem_end - self.elem_start - 1) / step + 1
             else:
                 xlen = (self.elem_start - self.elem_end - 1) / step + 1
+
+            if (self.elem_step < 0 and index > self.elem_start) or (
+                self.elem_step > 0 and index >= self.elem_end):
```
``` python
if self.elem_step < 0 and not (self.elem_end < index <= self.elem_start):
    raise IndexError(index)
elif self.elem_step > 0 and not (self.elem_start <= index < self.elem_end):
    raise IndexError(index)
```

klaussfreire (changes requested):

tests/mapped_struct.py
```diff
@@ -1500,6 +1500,27 @@ def testProxiedListRepr(self):
...
+        self.assertEquals(obj[::1], orig[::1])
+        self.assertEquals(obj[::2], orig[::2])
+        self.assertEquals(obj[::-1], orig[::-1])
+        self.assertEquals(obj[1::-1], orig[1::-1])
+        self.assertEquals(obj[2:][:5], orig[2:][:5])
+        self.assertEquals(obj[::3][::2], orig[::3][::2])
+        self.assertEquals(obj[1:10:2][2::5], orig[1:10:2][2::5])
+        self.assertEquals(obj[10:1:-3][2::2], orig[10:1:-3][2::2])
+        self.assertEquals(obj[huge:huge], orig[huge:huge])
+        self.assertEquals(obj[huge:huge:huge], orig[huge:huge:huge])
```
klaussfreire, line 1523: Add tests that test out of bounds access on slices (ie, the IndexError s). This equals thing will only test within-bounds tests.

Also add some "NotEquals" assertions to make sure the test fails if somehow the comparator is bogus.

klaussfreire (changes requested):

tests/mapped_struct.py
klaussfreire: You're missing the NotEqual

tests/mapped_struct.py
```diff
@@ -1534,6 +1534,8 @@ def testProxiedListSliceNotComparable(self):
         obj = self.pack(range(3))
         self.assertRaises(NotImplementedError, lambda: obj < "hello")
         self.assertRaises(NotImplementedError, lambda: obj >= 42)
+        self.assertNotEquals(obj, None)
+        self.assertNotEquals(obj, "123456")
```
klaussfreire, line 1541: Also not so bogus. Like not (obj == [1,2,4]). Etc. Be thorough.

klaussfreire (approved):

tests/mapped_struct.py
gefarion: @llogiudice I would add the corner cases: `[:]` and `[::]`.

gefarion: Maybe you can add some cases with things that are not valid indexes, like `[1:'a']` or `[1.5:4]`.

sharedbuffers/mapped_struct.py
```diff
@@ -812,13 +825,35 @@ def unpack_from(cls, buf, offs, idmap = None):
...
 
-    @cython.locals(obj_offs = cython.ulonglong, dcode = cython.char, index = cython.ulonglong, pindex = "unsigned long *", dataoffs = cython.ulonglong)
-    def __getitem__(self, index):
+    @cython.locals(obj_offs = cython.ulonglong, dcode = cython.char, index = cython.longlong, pindex = "unsigned long *",
+        dataoffs = cython.ulonglong, objlen = cython.longlong, xlen = cython.longlong, step = cython.longlong)
+    def _getitem(self, index):
 
         dcode, objlen, itemsize, dataoffs, _struct = self._metadata()
+        xlen = objlen
+        step = 1
```
gefarion: I think that this value is never used.

gefarion (approved):

gefarion (approved):

PR #53, opened 2018-10-23
Object pools and safer/better ID maps

klaussfreire:
This PR implements a bunch of machinery needed to incrementally construct large fully-connected object graphs.

The problem with fully connected object graphs is that they can't be constructed with the current API without holding them fully in memory. If they're big enough, that can be prohibitive.

This PR allows one to construct proxies into a dynamically growing staging area, replacing in-memory data structures with memory-mapped proxies. This releases some memory pressure by allowing the proxies to take the place of memory-intensive objects within the object graph. Finally, the contents of the dynamic buffers can be dumped together with any remaining objects by proper manipulation of id maps.

To make all this truly work, a lot of changes to the id mapping mechanism had to be introduced. An id map class that automatically discards transient objects was introduced. This is used as the default id map when constructing shared buffers now, since it's far safer than the old default. It's less efficient at object deduplication, however. To gain the old efficiency a set of "stable" objects (ones that we know are going to have a unique id for the duration of the packing procedure) can be specified when constructing the id map.

Furthermore, the id map structure was turned a bit more into a white box by exposing a pair of functions to construct ids: `shared_id` and `wrapped_id`. They can be used to pre-populate id maps with manually packed objects.

Additionally, the derivation of identity keys was changed to accomodate proxies in lieu of actual classes. Now proxy classes use their buffer-offset as identity key, whereas commonly used constants use their value, allowing automatic deduplication of commonly occurring primitive values across the board.

Some bugs in the existing id mapping code were fixed as well as fixes to some collections, like the ObjectIdMapper, that arose during testing.

### DIFF
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -4786,8 +4786,6 @@ class ObjectIdMapper(_CZipMapBase):
         stride0 = cython.size_t, stride1 = cython.size_t,
         indexbuf = 'Py_buffer', pindex = cython.p_char)
     def get(self, key, default = None):
-        if not isinstance(key, basestring):
-            return default
         hkey = _stable_hash(key)
         startpos = self._search_hkey(hkey)
         nitems = self.index_elements
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -123,6 +123,34 @@ class StrongIdMap(object):
         self.strong_refs = {}
         self.objmap = weakref.WeakValueDictionary()
 
+    def __len__(self):
+        return len(self.idmap)
+
+    def __iter__(self):
+        for key in self.idmap:
+            if key in self:
+                yield key
+
+    def iterkeys(self):
+        return iter(self)
+
+    def itervalues(self):
+        for key in self:
+            yield self[key]
+
+    def iteritems(self):
+        for key in self:
+            yield key, self[key]
+
+    def keys(self):
+        return list(self)
+
+    def values(self):
+        return list(self.itervalues())
+
+    def items(self):
+        return list(self.iteritems())
+
     def __setitem__(self, key, value):
         self.idmap[key] = value
         self.objmap[key] = NONE
@@ -256,7 +284,7 @@ class mapped_frozenset(frozenset):
 
 class mapped_tuple(tuple):
     @classmethod
-    @cython.locals(strong_refs = list, widmap = StrongIdMap)
+    @cython.locals(widmap = StrongIdMap)
     def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0,
             array = array.array):
         all_int = 1
@@ -535,7 +563,7 @@ class proxied_dict(object):
         iobuf = BufferIO(buf, offs)
         offs += ObjectIdMapper.build(_enum_keys(obj), iobuf, return_mapper=False)
         packer.pack_into(buf, ipos, iobuf.tell())
-        return proxied_list.pack_into(obj.values(), buf, offs, idmap, implicit_offs)
+        return proxied_list.pack_into([obj[k] for k in obj.iterkeys()], buf, offs, idmap, implicit_offs)
 
     @classmethod
     def unpack_from(cls, buf, offs, idmap = None):
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -119,14 +119,16 @@ class NONE:
 class StrongIdMap(object):
     cython.declare(
         __weakref__ = object,
+        preloaded = dict,
         idmap = dict,
         objmap = object,
         strong_refs = object,
     )
 
-    def __init__(self, strong_limit = 128 << 20, preallocate = False):
+    def __init__(self, strong_limit = 1 << 20, preallocate = False, strong_class = FastCache):
+        self.preloaded = {}
         self.idmap = {}
-        self.strong_refs = FastCache(
+        self.strong_refs = strong_class(
             strong_limit,
             eviction_callback = functools.partial(
                 self._on_strong_evict,
@@ -144,11 +146,13 @@ class StrongIdMap(object):
             self.pop(key, None)
 
     def __len__(self):
-        return len(self.idmap)
+        return len(self.idmap) + len(self.preloaded)
 
     def __iter__(self):
+        for key in self.preloaded:
+            yield key
         for key in self.idmap:
-            if key in self:
+            if key in self and key not in self.preloaded:
                 yield key
 
     def iterkeys(self):
@@ -176,6 +180,8 @@ class StrongIdMap(object):
         self.objmap[key] = NONE
 
     def __getitem__(self, key):
+        if key in self.preloaded:
+            return self.preloaded[key]
         if key not in self.objmap:
             self.idmap.pop(key, None)
             raise KeyError(key)
@@ -187,14 +193,22 @@ class StrongIdMap(object):
         del self.idmap[key]
 
     def __contains__(self, key):
-        return key in self.idmap and key in self.objmap
+        return key in self.preloaded or key in self.idmap and key in self.objmap
+
+    def preload(self, mapping):
+        self.preloaded.update(mapping)
 
     def clear(self):
         self.objmap.clear()
         self.idmap.clear()
         self.strong_refs.clear()
 
+    def clear_preloaded(self):
+        self.preloaded.clear()
+
     def get(self, key, default=None):
+        if key in self.preloaded:
+            return self.preloaded[key]
         if key not in self.objmap:
             self.idmap.pop(key, None)
             return default
@@ -202,6 +216,9 @@ class StrongIdMap(object):
 
     @cython.ccall
     def pop(self, key, default=NONE):
+        if key in self.preloaded:
+            # No popping from preloaded items
+            return self.preloaded[key]
         orv = self.objmap.pop(key, NONE)
         rv = self.idmap.pop(key, NONE)
         self.strong_refs.pop(key, None)
@@ -221,6 +238,9 @@ class StrongIdMap(object):
             # There's no need to link lifecycles, the object
             # is identified uniquely by its buffer mapping and that is a stable id
             return True
+        if is_equality_key(key):
+            # These are equality keys, they are hard references already
+            return True
 
         try:
             self.objmap[key] = obj
@@ -230,6 +250,14 @@ class StrongIdMap(object):
             # its id from being reused while we hold its idmap entry
             self.strong_refs[key] = obj
             return False
+
+
+@cython.ccall
+@cython.inline
+@cython.returns(cython.bint)
+def is_equality_key(obj):
+    # singletons and small strings
+    return obj is None or obj is () or (isinstance(obj, basestring) and len(obj) < 16)
 
 
 @cython.ccall
@@ -245,9 +273,33 @@ def shared_id(obj):
     elif isinstance(obj, BufferProxyObject):
         oobj = obj
         return (id(oobj.buf) << 64) | oobj.offs
+    elif is_equality_key(obj):
+        # Try an equality key, use the object itself as key if hashable
+        try:
+            hash(obj)
+        except TypeError:
+            pass
+        else:
+            return obj
+
+    # Otherwise plainly the id of the object
+    return id(obj)
+
+
+class WRAPPED:
+    pass
+
+
+@cython.ccall
+def wrapped_id(obj):
+    xid = shared_id(obj)
+    if isinstance(xid, (int, long)):
+        # We keep numeric ids numeric
+        xid |= 0xFL << 64
     else:
-        # Otherwise plainly the id of the object
-        return id(obj)
+        # Equality ids are wrapped in a tuple
+        xid = (WRAPPED, xid)
+    return xid
 
 
 class mapped_frozenset(frozenset):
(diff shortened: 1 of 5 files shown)


sharedbuffers/mapped_struct.py
```diff
@@ -106,8 +112,256 @@ def _likerobuffer(buf):
...
+    elif isinstance(obj, proxied_dict):
+        dobj = obj
+        return (id(dobj.buf) << 64) | dobj.offs
+    elif isinstance(obj, BufferProxyObject):
+        oobj = obj
+        return (id(oobj.buf) << 64) | oobj.offs
+    elif is_equality_key(obj):
+        # Try an equality key, use the object itself as key if hashable
+        try:
+            hash(obj)
```
llogiudice, line 321: Shouldn't this take into account the fact that python's hash isn't stable ?

sharedbuffers/mapped_struct.py
```diff
@@ -106,8 +112,256 @@ def _likerobuffer(buf):
...
+        # Equality ids are wrapped in a tuple
+        xid = (WRAPPED, xid)
+    return xid
+
+
+@cython.ccall
+@cython.inline
+@cython.returns(cython.bint)
+def is_wrapped_key(obj):
+    # singletons and small strings
```
llogiudice: This doesn't seem to handle small strings (Although it should, I think).

sharedbuffers/pool.py
```diff
@@ -0,0 +1,175 @@
...
+        else:
+            return 0
+
+    def add_section(self):
+        f = self._mktemp()
+        try:
+            f.truncate(self.section_size)
+            buf = mmap.mmap(
+                f.fileno(), 0,
+                access = mmap.ACCESS_WRITE)
```
llogiudice, line 74: Is this buffer going to be shared among processes ? If so, you need to call ```fsync``` periodically to make sure memory state is correctly synchronized.

sharedbuffers/pool.py
```diff
@@ -0,0 +1,175 @@
...
+
+import tempfile
+import mmap
+import struct
+
+from .mapped_struct import StrongIdMap
+
+# Default section size is set to 128MB which is a size at which most
+# malloc implementations turn to mmap
+DEFAULT_SECTION_SIZE = 128<<20
```
llogiudice, line 12: According to __man mallopt__, this threshold is at 128**KB**. On the other hand, it appears such threshold is dynamic, although it can be controlled with ```mallopt```.

sharedbuffers/mapped_struct.py
klaussfreire: I don't use the value of the hash, I just use that to test whether the object is hashable

sharedbuffers/mapped_struct.py
klaussfreire: I think that's a copypasto

sharedbuffers/pool.py
klaussfreire: Memory is always synchronized. `msync` (not fsync) would be needed to make sure the changes hit the disk. We don't currently require this.

sharedbuffers/pool.py
klaussfreire: It's dynamic, and the top limit for how big that dynamic threshold can get is 64MB.

klaussfreire: By default, of course

sharedbuffers/pool.py
```diff
@@ -0,0 +1,175 @@
...
+            new_section.fileobj = f
+
+            # Initialize with preloaded items
+            if self.idmap_preload:
+                idmap = {}
+                try:
+                    for schema, obj in self.idmap_preload:
+                        self._pack_into(schema, obj, new_section, None, idmap)
+                except (struct.error, IndexError):
+                    raise RuntimeError("Preload items dont't fit in empty section, increase section size")
```
llogiudice, line 88: Wouldn't it be possible to automatically readjust the section size upon failure ? Or does the caller expect the section size not to change ?

klaussfreire: Are you referring to an exponential growth strategy or something like that? It's already quite big at 128MB, I don't see much benefit from that.

I've thought about enlarging the section by truncating it to a bigger size, but the problem with that is that it would invalidate existing buffers.

llogiudice: Well, it doesn't necessarily have to be exponential, I think something like increasing by 50% on each retry should suffice.

The reason I asked about it is that sharedbuffers can already handle huge data structures gracefully by proxies, and this kind of error doesn't seem like a fatal exception, since you are in control of the temporary file and its size. I guess it's a secondary concern, though.

klaussfreire: Yes, but the caller is in control of both the segment size and the preloaded set. If the caller expects to preload something on all sections that fills up the section size it requested, it's already in deep trouble. Remember this goes on every section that is initialized, so if you're preloading that much data, you need to account for that when setting the section size, or you'll have tons of copies of the same huge data set.

So... I guess it's best if the caller knows about this, rather than "handling" it silently. This clearly is something the caller would not expect, or a bigger section size would have been requested.

klaussfreire: For context, this preload set is a set of objects you preload on all sections and their id maps, to be used as reference by the objects you're planning to add to the section. That's why it's unreasonable to expect the preload set to fill the section, it wouldn't leave any room for useful objects (those you add later with .pack) before another section with another copy of the preload set has to be allocated.

klaussfreire:
Ping

sharedbuffers/mapped_struct.py
```diff
@@ -106,8 +111,257 @@ def _likerobuffer(buf):
...
+
+    @staticmethod
+    @cython.locals(self = 'StrongIdMap')
+    def _on_strong_evict(wself, key, value):
+        self = wself()
+        if self is not None:
+            self.pop(key, None)
+
+    def __len__(self):
+        return len(self.idmap) + len(self.preloaded)
```
llogiudice, line 174: is this actually accurate ? based on the iter protocol below, it appears that the actual length would be the length of the intersection between the idmap and the preloaded map. Then again, it's possible that we don't really care about this length being 100% accurate.

klaussfreire: Yeah, it's not accurate, but it's not worth computing the intersection to get an accurate length either, IMO. Depends on what we use the len for, currently, approximate is enough (it's for logging and stuff like that)

sharedbuffers/mapped_struct.py
```diff
@@ -106,8 +111,263 @@ def _likerobuffer(buf):
     else:
         return buffer(buf)
 
+class NONE:
+    pass
+
+@cython.cclass
+class StrongIdMap(object):
```
tzulberti-jampp, line 118: Can't we move this to another file instead of adding even more classes to mapped_struct.py?

sharedbuffers/mapped_struct.py
```diff
@@ -106,8 +111,263 @@ def _likerobuffer(buf):
...
+            self.pop(key, None)
+
+    def __len__(self):
+        return len(self.idmap) + len(self.preloaded)
+
+    def __iter__(self):
+        for key in self.preloaded:
+            yield key
+        for key in self.idmap:
+            if key in self and key not in self.preloaded:
```
tzulberti-jampp, line 180: Why checking if `key in self` and not just checking `key in self.objmap`. Becuase the contains will check on on the idmap and we already know that is true.

sharedbuffers/mapped_struct.py
```diff
@@ -106,8 +111,263 @@ def _likerobuffer(buf):
...
+            # These are equality keys, they are hard references already
+            return True
+        if self.stable_set is not None:
+            if key in self.stable_set:
+                return True
+            elif is_wrapped_key(key) and get_wrapped_key(key) in self.stable_set:
+                return True
+
+        try:
+            self.objmap[key] = obj
```
tzulberti-jampp, line 277: I might be missing something if this object is added to the objmap, then the `get` is going to raise KeyError because it wasn't added to the `idmap`?

sharedbuffers/mapped_struct.py
```diff
@@ -106,8 +111,263 @@ def _likerobuffer(buf):
...
+
+@cython.ccall
+@cython.inline
+@cython.returns(cython.bint)
+def is_wrapped_key(obj):
+    # keys for type-tagged objects
+    if isinstance(obj, tuple):
+        return len(obj) == 2 and obj[0] is WRAPPED
+    elif isinstance(obj, (int, long)):
+        return (obj >> 64) == 0xF
```
tzulberti-jampp, line 355: Why doing this transformations for int/long? Just to check on `get_wrapped_key`?

sharedbuffers/mapped_struct.py
```diff
@@ -315,8 +609,14 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, itemsizes = {
...
             rv = list(array(dtype, buf[offs:offs+itemsizes[dtype]*objlen]))
-        elif dcode == 't':
-            dtype = 'l'
+        elif dcode == 't' or dcode == 'T':
+            if dcode == 't':
+                dtype = 'l'
+            elif dcode == 'T':
+                dtype = 'i'
+            else:
+                raise ValueError("Inconsistent data, unknown type code %r" % (dcode,))
```
tzulberti-jampp, line 618: You are already checking that dcode is `t` or `T`. Why adding this extra exception?

sharedbuffers/pool.py
```diff
@@ -0,0 +1,209 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
```
tzulberti-jampp, line 2: Aren't you going to cythinize this file?

sharedbuffers/mapped_struct.py
klaussfreire: This is not the PR to do such refactoring. Also, not until we bump the requirements to a modern Cython version so we can do proper refactoring without a performance hit.

sharedbuffers/mapped_struct.py
klaussfreire: It's easier to maintain. Like this, this code is correct no matter what happens in `__contains__`

klaussfreire: Also, the loop isn't atomic, so items can disappear mid-iteration.

sharedbuffers/mapped_struct.py
klaussfreire: The objmap just links the lifetimes of the idmap entry to that of the object. It's a way of detecting stale references. `link` doesn't add items to the idmap, it only links entries to objects, if the entry is there.

sharedbuffers/mapped_struct.py
klaussfreire: What transformation?

sharedbuffers/mapped_struct.py
klaussfreire: Just in case. Code changes.

sharedbuffers/pool.py
klaussfreire: I don't see the need for it ATM

llogiudice (approved):

sharedbuffers/mapped_struct.py
```diff
@@ -106,8 +111,263 @@ def _likerobuffer(buf):
...
+@cython.ccall
+@cython.locals(lobj = proxied_list, dobj = proxied_dict, oobj = BufferProxyObject)
+def shared_id(obj):
+    # (id(buf), offs) of buffer-mapped proxies
+    if isinstance(obj, proxied_list):
+        lobj = obj
+        rv = (id(lobj.buf) << 64) | lobj.offs
+        if lobj.elem_step != 0:
+            # Add slice arguments to the shared_id
+            rv |= long(lobj.elem_step) << (68 + 64)
```
gefarion, line 308: Why 68 bits? It is not 64 + 64?

klaussfreire: 64 + 4 bits for flags

gefarion (approved):

klaussfreire:
Only @tzulberti-jampp remaining...

sharedbuffers/pool.py
```diff
@@ -0,0 +1,209 @@
...
+            size = len(self.buf) - write_pos
+        if (write_pos + size) <= len(self.buf):
+            self.write_pos += size
+            return write_pos
+        else:
+            raise IndexError("Buffer overflow trying to allocate %d bytes from section" % size)
+
+    def resize(self, pos, new_size):
+        new_end = pos + new_size
+        if new_end != self.write_pos:
```
tzulberti-jampp: I wasn't able to find an example of this function call, so I am not sure about this, but should be that pos == self.write_pos? Why checking the new_end?

sharedbuffers/pool.py
```diff
@@ -0,0 +1,209 @@
...
+    def clear_idmaps(self):
+        for section in self.sections:
+            section.idmap.clear()
+
+    def dump(self, fileobj):
+        if not self.sections:
+            return 0
+
+        for section in self.sections[:-1]:
+            write_bytes = len(section.real_buf)
```
tzulberti-jampp: Is this variable being used?

sharedbuffers/pool.py
klaussfreire: I meant to use it, but I didn't. The intention was to be able to "pack_into" the mmap directly, allocating a generous buffer, and then resize that buffer. But as you say, the old_end should be the one being compared, and I don't have it in this context. Since we didn't use it in the end, because packing directly to the mmap resulted impossible due to mmap module limitations, I'm going to scrap the whole method and call it a day.

sharedbuffers/pool.py
klaussfreire: Nope

tzulberti-jampp (approved):

PR #54, opened 2018-10-26
Fix buffer leak in proxies

klaussfreire:
Cython doesn't invoke __del__ on extension types. Also,
Py_buffer should be initialized in cinit.


llogiudice (approved):

tzulberti-jampp (commented):
Shouldn't this PR be included on v0.4.9?

tzulberti-jampp (approved):

klaussfreire:
We can backport the fix later and release 0.4.9 if necessary.

PR #56, opened 2018-10-29
Buffer leak fix backport

klaussfreire:


sharedbuffers/mapped_struct.py
```diff
@@ -683,18 +683,21 @@ class BufferProxyObject(object):
         none_bitmap = cython.ulonglong
     )
 
-    @cython.locals(offs = cython.ulonglong, none_bitmap = cython.ulonglong)
-    def __init__(self, buf, offs, none_bitmap, idmap = None):
+    def __cinit__(self, buf, offs, none_bitmap, idmap = None):
```
tzulberti-jampp, line 686: Shouldn't this be calling `_init`. Also you already know that in this case, the`cython.compiled` is going to be true

klaussfreire: Not really. In pure python people could manually call `obj.__cinit__`, and I wouldn't want weird errors if they did

klaussfreire: Also, not, `__init__` will still be called, and that will call `_init`.

llogiudice (approved):

tzulberti-jampp (approved):

PR #57, opened 2018-11-07
Pack optimizations

tzulberti-jampp:


sharedbuffers/mapped_struct.py
```diff
@@ -332,12 +361,18 @@ class WRAPPED:
     pass
 
 
+WRAP_MASK = cython.declare(object, 1L << 64)
```
tzulberti-jampp, line 364: Are this constants defined here for any particular reason?

sharedbuffers/mapped_struct.py
```diff
@@ -332,12 +361,18 @@ class WRAPPED:
...
+FLAG_MASK = cython.declare(object, 0xFL << 64)
+
+
 @cython.ccall
 def wrapped_id(obj):
     xid = shared_id(obj)
     if isinstance(xid, (int, long)):
         # We keep numeric ids numeric
-        xid |= 0xFL << 64
+        xid |= WRAP_MASK
```
tzulberti-jampp, line 375: Shouldn't this be FLAG_MASK? or there was a bug about that?

sharedbuffers/mapped_struct.py
```diff
@@ -441,72 +511,132 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+                        buf[offs] = 'q'
+                    elif dtype == 'L':
+                        buf[offs] = 'Q'
+                    else:
+                        buf[offs] = dtype
+        else:
+            all_int = all_intlong = all_float = 1
+            minval = maxval = 0
+            for x in obj:
+                if all_int and type(x) is not int:
```
tzulberti-jampp, line 557: Is this faster than checking `isinstance(x, int)`?

tzulberti-jampp, line 557: Why don't just calculate the type once: `x_type = type(x)`?

sharedbuffers/mapped_struct.py
klaussfreire: To name them

sharedbuffers/mapped_struct.py
klaussfreire: Nope. FLAG_MASK is a mask to get all flags, WRAP_MASK is the "wrapped" flag only

sharedbuffers/mapped_struct.py
klaussfreire: Yes, considerably. Also, not exactly the same. This doesn't accept subclasses, and in many cases we don't want to accept subclasses.

klaussfreire: It would be counterproductive. Check the generated code.

PR #58, opened 2018-11-14
Pack optimizations

klaussfreire:
Tons of optimizations, some format changes too

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -49,7 +49,7 @@ Changes
 - New wide bitmap frozenset format allows more frozensets to be packed
   as bitmaps.
 - Offsets are Py_ssize_t now. That shouldn't be a noticeable change,
-  unless you've got more storage than they universe.
+  unless you've got more storage than the universe.
 
 Bugfixes
 --------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -46,6 +46,8 @@ Changes
   containers though.
 - Shrink some buffers by employing narrow pointers where possible
 - Optimized sequence packing
+- New wide bitmap frozenset format allows more frozensets to be packed
+  as bitmaps.
 
 Bugfixes
 --------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -48,6 +48,8 @@ Changes
 - Optimized sequence packing
 - New wide bitmap frozenset format allows more frozensets to be packed
   as bitmaps.
+- Offsets are Py_ssize_t now. That shouldn't be a noticeable change,
+  unless you've got more storage than they universe.
 
 Bugfixes
 --------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -45,6 +45,7 @@ Changes
   perhaps considerably. Proper identity was implemented for proxied
   containers though.
 - Shrink some buffers by employing narrow pointers where possible
+- Optimized sequence packing
 
 v0.4.8 - 2018-05-28
 ===================
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -411,7 +411,8 @@ class mapped_frozenset(frozenset):
             obj_dtype = obj.dtype
             if obj_dtype.isbuiltin:
                 dtype = obj_dtype.char
-                if dtype in ('L', 'l','I','i','H','h','B','b'):
+                cdtype = cython.cast('const char*', dtype)[0]
+                if cdtype in ('L', 'l', 'I', 'i', 'H', 'h', 'B', 'b'):
                     all_int = 1
         else:
             all_int = 1
@@ -533,20 +534,21 @@ class mapped_tuple(tuple):
             obj_dtype = obj.dtype
             if obj_dtype.isbuiltin:
                 dtype = obj_dtype.char
-                if dtype in ('l','I','i','H','h','B','b'):
+                cdtype = cython.cast('const char*', dtype)[0]
+                if cdtype in ('l', 'I', 'i', 'H', 'h', 'B', 'b'):
                     all_int = all_intlong = 1
                     all_float = 0
-                elif dtype == 'L':
+                elif cdtype == 'L':
                     all_int = all_float = 0
                     all_intlong = 1
-                elif dtype in ('d','f'):
+                elif cdtype in ('d', 'f'):
                     all_int = all_intlong = 0
                     all_float = 1
                 if all_int or all_intlong or all_float:
                     # translate l -> q
-                    if dtype == 'l':
+                    if cdtype == 'l':
                         buf[offs] = 'q'
-                    elif dtype == 'L':
+                    elif cdtype == 'L':
                         buf[offs] = 'Q'
                     else:
                         buf[offs] = dtype
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -2880,18 +2880,29 @@ class Schema(object):
 
     @cython.ccall
     @cython.locals(has_bitmap = cython.ulonglong, none_bitmap = cython.ulonglong, present_bitmap = cython.ulonglong,
-        i = int)
+        i = int, get_values = cython.bint)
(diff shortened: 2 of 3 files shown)


sharedbuffers/mapped_struct.py
```diff
@@ -421,12 +484,19 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+            fs_type = pbuf[offs]
+            if fs_type == 'm' or fs_type == 'M':
                 # inline bitmap
-                if cython.compiled and offs+7 >= pybuf.len:
+                if fs_type == 'm':
+                    fs_size = 7
+                elif fs_type == 'M':
+                    fs_size = 15
+                else:
+                    raise ValueError("Unknown set type %r" % fs_type)
```
llogiudice, line 496: isn't this check a bit unnecessary, since you've already tested for those types above ?

sharedbuffers/mapped_struct.py
```diff
@@ -2667,18 +2904,29 @@ def init(self, slot_types = None, slot_keys = None, alignment = 8, pack_buffer_s
...
+    def _get_bitmaps(self, obj, get_values=False):
         has_bitmap = 0
         none_bitmap = 0
+        if get_values:
+            values = []
         for i,slot in enumerate(self.slot_keys):
-            if hasattr(obj, slot):
+            try:
+                val = getattr(obj, slot)
+            except AttributeError:
```
llogiudice, line 2939: is this actually better than doing a ```hasattr``` test ? I'd have thought that exception handling was always slower.

sharedbuffers/mapped_struct.py
```diff
@@ -2667,18 +2904,29 @@ def init(self, slot_types = None, slot_keys = None, alignment = 8, pack_buffer_s
...
+                if val is None:
                     none_bitmap |= cython.cast(cython.ulonglong, 1) << i
+                elif get_values:
+                    values.append(val)
         present_bitmap = has_bitmap & ~none_bitmap
-        return has_bitmap, none_bitmap, present_bitmap
+        if get_values:
+            return has_bitmap, none_bitmap, present_bitmap, values
+        else:
+            return has_bitmap, none_bitmap, present_bitmap
```
llogiudice, line 2951: shouldn't this return a list in its fourth return value so the unpacking in ```get_packable``` doesn't fail ?

sharedbuffers/mapped_struct.py
klaussfreire: Yes, but it's free and it may save your life if the code that makes it unnecessary changes

sharedbuffers/mapped_struct.py
klaussfreire: Exception handling in Cython is quite fast. That said, I haven't timed it on basic attributes. But it's most likely faster due to another subtle reason: hasattr on properties (ie: all proxied properties) has to invoke the getter and check if it rases AttributeError. That is, this exact code, only it discards the value if it succeeds. This is faster on properties because it avoids one whole invocation to the getter, which can be quite expensive.

sharedbuffers/mapped_struct.py
klaussfreire: The whole point of `get_values` is to change the shape of the returned tuple

tests/mapped_struct.py
```diff
@@ -336,8 +336,8 @@ def testUnpackInto(self):
             px = self.schema.pack(x)
             dx = self.schema.unpack(px, proxy_into = dx)
             for k,v in TEST_VALUES.iteritems():
-                self.assertTrue(hasattr(dx, k))
                 self.assertEqual(getattr(dx, k), v)
+                self.assertTrue(hasattr(dx, k))
```
tzulberti-jampp: Why did you change the order of this? It should check that it has the attribute before checking the value

sharedbuffers/mapped_struct.py
```diff
@@ -441,72 +511,132 @@ def unpack_from(cls, buf, offs, idmap = None):
...
-                buf[offs] = 'q'
-                dtype = 'l'
-            if dtype == 'l':
-                buf[offs+1:offs+8] = struct.pack('<Q', objlen)[:7]
+        if isinstance(obj, npndarray):
+            all_int = all_intlong = all_float = 0
+            obj_dtype = obj.dtype
+            if obj_dtype.isbuiltin:
+                dtype = obj_dtype.char
+                if dtype in ('l','I','i','H','h','B','b'):
```
tzulberti-jampp: nit: space between the tuple values?

sharedbuffers/mapped_struct.py
```diff
@@ -532,29 +664,77 @@ def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0,
             else:
                 widmap = None
 
+            # Get a sense of whether we can use narrow pointers
+            min_offs = max_offs = 0
+            for i,x in enumerate(obj):
```
tzulberti-jampp: nit: space between `i,x`

tests/mapped_struct.py
klaussfreire: I was doing some tests and forgot to revert that

CHANGELOG.rst
```diff
@@ -45,6 +45,11 @@ Changes
   perhaps considerably. Proper identity was implemented for proxied
   containers though.
 - Shrink some buffers by employing narrow pointers where possible
+- Optimized sequence packing
+- New wide bitmap frozenset format allows more frozensets to be packed
+  as bitmaps.
+- Offsets are Py_ssize_t now. That shouldn't be a noticeable change,
+  unless you've got more storage than they universe.
```
tzulberti-jampp: Shouldn't it be `than the universe`?

sharedbuffers/mapped_struct.py
```diff
@@ -421,12 +485,19 @@ def unpack_from(cls, buf, offs, idmap = None):
...
         try:
-            if pbuf[offs] == 'm':
+            fs_type = pbuf[offs]
+            if fs_type == 'm' or fs_type == 'M':
                 # inline bitmap
-                if cython.compiled and offs+7 >= pybuf.len:
+                if fs_type == 'm':
+                    fs_size = 7
+                elif fs_type == 'M':
+                    fs_size = 15
```
tzulberti-jampp, line 494: Shouldn't this be 14? Because the for in line 500 is adding a +1.

klaussfreire: No, `M` has 8 bytes more than `m`. The loop in 500 will go all the way up to i=14, which means `pbuf[offs+1+14] == pbuf[offs+15]` which is the last byte of a 16-byte bitmap.

klaussfreire:
I'm going to stop adding stuff in here to incentivize reviews.

So... @tzulberti-jampp , @gefarion , @llogiudice , @npazos , it's your ball now

klaussfreire:
Ping

sharedbuffers/mapped_struct.py
```diff
@@ -385,26 +435,40 @@ def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0):
...
                         cbuf[offs+1+ix/8] |= 1 << (ix & 7)
                 else:
-                    buf[offs:offs+1] = 'm'
-                    buf[offs+1:offs+8] = '\x00\x00\x00\x00\x00\x00\x00'
+                    if maxval < 56:
+                        buf[offs:offs+1] = 'm'
+                        buf[offs+1:offs+8] = '\x00\x00\x00\x00\x00\x00\x00'
+                    else:
+                        buf[offs:offs+1] = 'M'
+                        buf[offs+1:offs+15] = '\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
```
llogiudice, line 454: I think there's a mismatch here.
When ```maxval < 56```, you write 7 bytes, using offsets 1 to 8
When ```maxval >= 56```, you write 15 bytes, using offsets 1 to 15

I think in the second case, you have to use offsets 1 to 16 (And also update offs accordingly below).

tzulberti-jampp (approved):

llogiudice (approved):

PR #59, opened 2018-11-14
Proxied frozenset

llogiudice:


sharedbuffers/mapped_struct.py
```diff
@@ -131,7 +131,7 @@ def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0):
                 return tup.pack_into(tup, buf, offs, idmap, implicit_offs)
         else:
             # Same representation as a tuple of items, only backed in-memory by a frozenset
-            tup = mapped_tuple(obj)
+            tup = mapped_tuple(obj if sort is False else sorted(obj))
```
klaussfreire: Sorting should be unconditional if feasible. The whole point is that we should be able to switch from regular frozensets to proxied ones without rebuilding the underlying representation (they should be compatible), and an unsorted frozenset wouldn't be compatible.

That said, you should try to make sorting as efficient as possible. Using python's "sorted" wouldn't achieve that. `mapped_tuple` has optimized paths for packing numpy arrays IIRC, sorted would break that. And numpy arrays can be sorted more efficiently with numpy itself.

sharedbuffers/mapped_struct.py
```diff
@@ -1001,6 +1001,166 @@ def __str__(self):
     setattr(proxied_list, '__eq__', getattr(proxied_list, '_eq'))
     setattr(proxied_list, '__ne__', getattr(proxied_list, '_ne'))
 
+@cython.cclass
+class proxied_frozenset(object):
```
klaussfreire, line 1756: Shouldn't you inherit from mapped_frozenset?

sharedbuffers/mapped_struct.py
```diff
@@ -1001,6 +1001,166 @@ def __str__(self):
...
+        while cnt > 0:
+            step = cnt >> 1
+            val = lst._getitem(step)
+            if val == elem:
+                return True
+            elif val < elem:
+                cnt -= step + 1
+            else:
+                cnt = step
+        return False
```
klaussfreire, line 1908: Just no. We talked about this. We care most about optimizing numeric sets, and we want numeric sets to be as fast as possible. This is contrary to that. Going over _getitem, and in that fashion, has considerable overhead. Especially when you know you can use search_hkey for numeric sets.

Since you *know* objlist is a proxied_list, you can reach for its `pybuf` and use it directly here, then pass pointers to search_hkey.

sharedbuffers/mapped_struct.py
```diff
@@ -1001,6 +1001,166 @@ def __str__(self):
...
+    def __len__(self):
+        return len(self.objlist)
+
+    @cython.locals(i1=cython.Py_ssize_t, i2=cython.Py_ssize_t,
+        cnt=cython.Py_ssize_t, step=cython.Py_ssize_t, lst='proxied_list')
+    def __contains__(self, elem):
+        i1 = 0
+        i2 = len(self.objlist)
+        cnt = i2 - i1
+        lst = self.objlist
```
klaussfreire: objlist = self.objlist.

Or simply self.objlist, a cython attribute can be accessed very efficiently.

sharedbuffers/mapped_struct.py
```diff
@@ -1001,6 +1001,166 @@ def __str__(self):
     setattr(proxied_list, '__eq__', getattr(proxied_list, '_eq'))
     setattr(proxied_list, '__ne__', getattr(proxied_list, '_ne'))
 
+@cython.cclass
+class proxied_frozenset(object):
+
+    cython.declare(objlist = object)
```
klaussfreire: Should be of type proxied_list

sharedbuffers/mapped_struct.py
```diff
@@ -1001,6 +1001,166 @@ def __str__(self):
...
+    @cython.locals(
+        i=int, j=int, offs=cython.longlong,
+        pybuf='Py_buffer', pbuf='const unsigned char *', b=cython.uchar)
+    def unpack_from(cls, buf, offs, idmap = None):
+        buf = _likerobuffer(buf)
+        if cython.compiled:
+            PyObject_GetBuffer(buf, cython.address(pybuf), PyBUF_SIMPLE)
+            pbuf = cython.cast(cython.p_uchar, pybuf.buf)
+            if offs >= pybuf.len:
+                PyBuffer_Release(cython.address(pybuf))
```
klaussfreire: The Release needs to be in a finally. Otherwise, an exception in any other place would leak a reference.

This means you could have just as easily put it inside the try/finall below.

sharedbuffers/mapped_struct.py
```diff
@@ -1001,6 +1001,166 @@ def __str__(self):
...
+                if cython.compiled and offs+7 >= pybuf.len:
+                    raise IndexError("Object spans beyond buffer end")
+                rv = []
+                for i in xrange(7):
+                    b = ord(pbuf[offs+1+i])
+                    if b:
+                        for j in xrange(8):
+                            if b & (1<<j):
+                                rv.append(i*8+j)
+                return frozenset(rv)
```
klaussfreire: You should create an object that takes the buffer and queries the bitmap for `__contains__` rather than fleshing out the frozenset. In essence, this isn't a proxied set. This is a regular mapped_frozenset here.

sharedbuffers/mapped_struct.py
```diff
@@ -1001,6 +1001,166 @@ def __str__(self):
...
+    def symmetric_difference(self, *seqs):
+        return frozenset(self).symmetric_difference(*seqs)
+
+    def __eq__(self, seq):
+        if not isinstance(seq, (set, frozenset, proxied_frozenset, mapped_frozenset)) or (
+              len(self) != len(seq)):
+            return False
+        for val in self:
+            if val not in seq:
+                return False
```
klaussfreire, line 2095: Actually, you can implement it far more efficiently if the other sequence is a proxied_frozenset as well, in O(N).

sharedbuffers/mapped_struct.py
```diff
@@ -1001,6 +1001,166 @@ def __str__(self):
...
+        return frozenset(self).difference(*seqs)
+
+    def symmetric_difference(self, *seqs):
+        return frozenset(self).symmetric_difference(*seqs)
+
+    def __eq__(self, seq):
+        if not isinstance(seq, (set, frozenset, proxied_frozenset, mapped_frozenset)) or (
+              len(self) != len(seq)):
+            return False
+        for val in self:
```
klaussfreire: Don't use naive iteration, it's slow. Reach for the buffer and iterate the buffer in an optimized fashion.

sharedbuffers/mapped_struct.py
```diff
@@ -1001,6 +1001,166 @@ def __str__(self):
...
+        return frozenset(self).intersection(*seqs)
+
+    def difference(self, *seqs):
+        return frozenset(self).difference(*seqs)
+
+    def symmetric_difference(self, *seqs):
+        return frozenset(self).symmetric_difference(*seqs)
+
+    def __eq__(self, seq):
+        if not isinstance(seq, (set, frozenset, proxied_frozenset, mapped_frozenset)) or (
```
klaussfreire: Beware, Cython does not guarantee that self will be a proxied_frozenset or that seq will be either. You know *one* of them will, so you have to implement a helper `__eq__` that assumes `self` is a proxied_frozenset, and check the type of arguments. Kinda like:

``` python
    _frozenset_eq(self, other)
else:
    _frozenset_eq(other, self)
```

Then you can have efficient code, since "self" will be of a known type.

sharedbuffers/mapped_struct.py
```diff
@@ -1001,6 +1001,166 @@ def __str__(self):
...
+              len(self) != len(seq)):
```
klaussfreire: Bad indentation

sharedbuffers/mapped_struct.py
```diff
@@ -1001,6 +1001,166 @@ def __str__(self):
...
+        if len(self) > len(seq):
+            return False
+
+        for val in self:
+            if val not in seq:
+                return False
+        return not strict_subset or len(self) < len(seq)
+
+    def __lt__(self, seq):
+        return self._subset(seq, True)
```
klaussfreire: Same here. Self won't always be of the right type. You have to make your code support both calling conventions `(proxied_frozenset, something_else)` and `(something_else, proxied_frozenset)`.

Just test it, try: `frozenset([1]) < proxied_frozenset([1,2])`, you'll see it blows up.

sharedbuffers/mapped_struct.py
```diff
@@ -1001,6 +1001,166 @@ def __str__(self):
...
+    @cython.ccall
+    @cython.locals(strict_subset=cython.bint)
+    def _subset(self, seq, strict_subset):
+        if len(self) > len(seq):
+            return False
+
+        for val in self:
+            if val not in seq:
+                return False
+        return not strict_subset or len(self) < len(seq)
```
klaussfreire: If both arguments are a proxied_frozenset, you can optimize this considerably, since you can do it in O(N) merge-style. You can probably use search_hkey's hints to get a merge without having to code it.

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
llogiudice: The input for frozenset construction can be an arbitrary sequence, so I'm not sure if we can sort unconditionally. I can apply a numpy-specific sorting since that's likely to be better.

sharedbuffers/mapped_struct.py
llogiudice: None of the proxy classes inherit from the mapped ones, so I didn't see the point in doing that.

sharedbuffers/mapped_struct.py
llogiudice: OK, I'll see what I can do. I did notice the search_hkey* functions, but I thought they used an index to reference entries.

sharedbuffers/mapped_struct.py
llogiudice: Got it.

sharedbuffers/mapped_struct.py
llogiudice: Right.

sharedbuffers/mapped_struct.py
llogiudice: I copied this from mapped_frozenset, I'll fix it.

sharedbuffers/mapped_struct.py
llogiudice: That's true. I did it that way because I felt that for such short frozensets, allocating a fresh one didn't make much of a difference.

sharedbuffers/mapped_struct.py
llogiudice: That is O(N), though.

It can be more efficient, though, I agree. I could just compare elements sequentially, or even do something like ```memcmp``` if the underlying typecodes are equal.

sharedbuffers/mapped_struct.py
llogiudice: Is it really that slow ? proxied_list's ```_getitem``` already fetches the buffer and dereferences the pointer. The only source of overhead I can think of is the tuple that is created by the ```_metadata``` method, but we can work around that.

sharedbuffers/mapped_struct.py
llogiudice: The test you proposed works fine for me ...

I can make the change, though. It's trivial.

tzulberti-jampp (commented):
Please update the CHANGELOG

sharedbuffers/mapped_struct.py
klaussfreire: Items are hashable, so they're sortable by hash. We talked about it.

sharedbuffers/mapped_struct.py
klaussfreire: It was an idea, since I expect you'll reuse a lot of the machinery there. But whatever.

sharedbuffers/mapped_struct.py
klaussfreire: I guess it depends, but since you've got access to the buffer already, I would expect instantiating a proxy would be less overhead than instantiating a full set. Also, lookups in a bitmap are O(1).

sharedbuffers/mapped_struct.py
klaussfreire: You're quite wrong. In more ways than one:

1. You're confusing _O(1)_ with _average O(1)_. Hash tables are O(N) worst case, and that's not an esoteric worst case, python's hashes were exploited and a DoS of web servers could be crafted by exploiting that.
2. You're assuming `seq` is a hash. It may not be. It could be another proxied_frozenset in which case each test would be _O(log N)_, so you've got _O(N log N)_
3. We don't only care about asymptotic complexity. The "constant factor" matters a lot, that's why we cythonize.

sharedbuffers/mapped_struct.py
klaussfreire: `for val in self` has to construct and iterate a generator.

Yes, it's very high overhead, even if we optimize the generator. At the very least you should inline the generator loop to avoid the generator indirection.

llogiudice: Good point. A C-style index iteration should be much faster. Thanks.

sharedbuffers/mapped_struct.py
```diff
@@ -1019,15 +1019,15 @@ def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0):
...
-            PyObject_GetBuffer(buf, cython.address(pybuf), PyBUF_SIMPLE)
-            pbuf = cython.cast(cython.p_uchar, pybuf.buf)
-            if offs >= pybuf.len:
-                PyBuffer_Release(cython.address(pybuf))
-                raise IndexError("Offset out of range")
-        else:
-            pbuf = buf
         try:
+            if cython.compiled:
+                PyObject_GetBuffer(buf, cython.address(pybuf), PyBUF_SIMPLE)
```
klaussfreire, line 1784: To avoid releasing an uninitialized buffer if obtaining the buffer itself fails, set `pybuf.buf = cython.NULL` before anything else, and check for NULL before releasing.

sharedbuffers/mapped_struct.py
```diff
@@ -1086,49 +1084,87 @@ def difference(self, *seqs):
...
+            return self.objlist == pfset.objlist
+
+        xlen = len(self)
+        if not isinstance(x, (set, frozenset)) or xlen != len(x):
             return False
-        for val in self:
-            if val not in seq:
+
+        i = 0
+        while i < xlen:
```
klaussfreire: This is a `for i in xrange(xlen)` (cython will generate the same code, or maybe better)

sharedbuffers/mapped_struct.py
```diff
@@ -1086,49 +1084,87 @@ def difference(self, *seqs):
...
                 return False
+            i += 1
         return True
 
+
+    def __eq__(self, seq):
+        if isinstance(self, proxied_frozenset):
+            return self._frozenset_eq(seq)
+        else:
+            return seq._frozenset_eq(self)
```
klaussfreire: You have to cast self/seq to `proxied_frozenset` for ccall to be effective

sharedbuffers/mapped_struct.py
```diff
@@ -1086,49 +1084,87 @@ def difference(self, *seqs):
...
+            pfset = cython.cast('proxied_frozenset', seq)
+            seqlen = len(pfset.objlist)
+            if xlen > seqlen:
+                return False
+
+            while i < xlen:
+                val = self.objlist._getitem(i)
+                while True:
+                    c = cmp(val, pfset.objlist._getitem(j))
+                    j += 1
```
klaussfreire: You have to implement the same exponential hopping search_hkey uses to avoid this being _O(max(N, M))_. That way, the complexity is closer to _O(N)_ (though not really, more like _O(N log (1+M-N))_, but still closer than otherwise).

In essence, instead of `j += 1`, you use an exponentially-growing step, and if you step too far, you move back and reset the step back to 1. Do use search_hkey as inspiration.

In fact, for numeric arrays, you can outright use search_hkey with a proper hint pointing to the last `j`.

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
llogiudice: I didn't know that. Thanks.

sharedbuffers/mapped_struct.py
llogiudice: Also wasn't aware of this.

sharedbuffers/mapped_struct.py
```diff
@@ -1001,6 +1001,321 @@ def __str__(self):
...
+            i += 1
+        x >>= 1
+    return i
+
+@cython.cclass
+class proxied_frozenset(object):
+
+    cython.declare(objlist = 'proxied_list',
+      bitrep_lo=cython.ulonglong, bitrep_hi=cython.ulonglong,
+      bitlen=cython.Py_ssize_t)
```
klaussfreire: Use proper indentation and be consistent with whitespace use

sharedbuffers/mapped_struct.py
```diff
@@ -1001,6 +1001,321 @@ def __str__(self):
...
+
+    def __init__(self, objlist, bitrep_lo=0, bitrep_hi=0):
+        self.objlist = objlist
+        self.bitrep_lo = bitrep_lo
+        self.bitrep_hi = bitrep_hi
+        self.bitlen = -1 if self.objlist is not None else (
+            _count_bits_set(self.bitrep_lo) + _count_bits_set(self.bitrep_hi))
```
klaussfreire: Why count now and not on-demand in `__len__`?

sharedbuffers/mapped_struct.py
```diff
@@ -1001,6 +1001,321 @@ def __str__(self):
...
+@cython.ccall
+@cython.inline
+@cython.locals(x=cython.ulonglong, i=cython.ulonglong)
+def _count_bits_set(x):
+    i = 0
+    while x != 0:
+        if (x & 1) != 0:
+            i += 1
+        x >>= 1
+    return i
```
klaussfreire: Check out [this stack overflow](https://stackoverflow.com/questions/9946115/hamming-weight-written-only-in-binary-operations), you've got quite a few interesting implementations there.

sharedbuffers/mapped_struct.py
```diff
@@ -1001,6 +1001,321 @@ def __str__(self):
...
+                if offs >= pybuf.len:
+                    raise IndexError("Offset out of range")
+            else:
+                pbuf = buf
+
+            if pbuf[offs] == 'm':
+                # inline bitmap (64 bits)
+                if cython.compiled and offs+7 >= pybuf.len:
+                    raise IndexError("Object spans beyond buffer end")
+                bitrep_lo = cython.cast(cython.p_ulonglong, pbuf + offs + 1)[0]
```
klaussfreire: Why are you using unaligned access?

Yo can simply read the typecode and shift it out:

``` python
bitrep_lo = cython.cast(cython.p_ulonglong, buf + offs)[0] >> 8
```

sharedbuffers/mapped_struct.py
```diff
@@ -1001,6 +1001,321 @@ def __str__(self):
...
+                return proxied_frozenset(None, bitrep_lo, 0)
+            elif pbuf[offs] == 'M':
+                # inline bitmap (128 bits)
+                if cython.compiled and offs+15 >= pybuf.len:
+                    raise IndexError("Object spans beyond buffer end")
+                bitrep_lo = cython.cast(cython.p_ulonglong, pbuf + offs + 1)[0]
+                bitrep_hi = cython.cast(cython.p_ulonglong, pbuf + offs + 1)[1]
```
klaussfreire: Similarly here:

Why are you using unaligned access?

Yo can simply read the typecode and shift it out:

``` python
bitrep_lo = cython.cast(cython.p_ulonglong, buf + offs)[0] >> 8
bitrep_hi = cython.cast(cython.p_ulonglong, buf + offs)[1]
bitrep_lo |= (bitrep_hi & 0xFF) << 56
bitrep_hi >>= 8
```

Note that you simply can't use unaligned access, because the last byte of that 64-bit ulonglong might be out of bounds.

sharedbuffers/mapped_struct.py
```diff
@@ -1076,19 +1077,43 @@ def __len__(self):
...
+        if dcode == 'q':
+            return _c_search_hkey_i64(elem, pindex, 8, xlen, hint)
+        elif dcode == 'I':
+            return _c_search_hkey_ui32(elem, pindex, 4, xlen, hint)
+        elif dcode == 'i':
+            return _c_search_hkey_i32(elem, pindex, 4, xlen, hint)
+        elif dcode == 'd':
+            return _c_search_hkey_f64(elem, pindex, 8, xlen, hint)
+        else:
+            raise NotImplementedError("invalid frozenset data type %s" % chr(dcode))
```
klaussfreire: You do have to implement it. You could use bisect, or you could implement the relevant _c_search_hkey variants. It's mostly copypaste work.

sharedbuffers/mapped_struct.py
```diff
@@ -1076,19 +1077,43 @@ def __len__(self):
...
                 return (eint < 64 and (self.bitrep_lo & (1 << eint)) != 0) or (
-                    (eint < 128 and (self.bitrep_hi & (1 << eint)) != 0))
+                    (eint < 128 and (self.bitrep_hi & (1 << (eint - 64))) != 0))
             except (ValueError, TypeError):
                 return False
 
+        if cython.compiled:
+            dcode, _, _, offset, _ = self.objlist._metadata()
+            if dcode not in ('H', 'h', 't'):
+                i1 = self._contains(elem, dcode, offset)
```
klaussfreire: Why doesn't _contains return a boolean?

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
llogiudice: No particular reason, but since ```len``` is used everywhere, I don't see much gain in computing it lazily.

sharedbuffers/mapped_struct.py
llogiudice: Yes, I tried to see if cython allowed to me to get to gcc's ```__builtin_popcount```, to no avail. On the other hand, this is only called twice, so I didn't see much gain in doing it this way.

sharedbuffers/mapped_struct.py
llogiudice: Good points.

sharedbuffers/mapped_struct.py
llogiudice: Oh, this will actually never be reached, as I check before calling this method. I guess I should change the exception's message.

sharedbuffers/mapped_struct.py
llogiudice: I was planning on using it in ```subset```, to implement skip search, as you suggested.

sharedbuffers/mapped_struct.py
klaussfreire: You can define it in the pxd:

``` python
cdef extern from *:
    cdef int __builtin_popcountll(unsigned long long x)
```

sharedbuffers/mapped_struct.py
```diff
@@ -1080,19 +1080,18 @@ def __len__(self):
...
-    def _contains(self, elem, dcode, offset):
-        xlen = len(self.objlist)
-        hint = xlen // 2
+    @cython.locals(dcode=cython.char, offset=cython.size_t,
+        start=cython.size_t, xlen=cython.size_t, hint=cython.size_t)
+    def _search_key(self, elem, dcode, offset, start, xlen, hint):
         pindex = cython.cast(cython.p_char, self.objlist.pybuf.buf) + offset
         if dcode == 'q':
-            return _c_search_hkey_i64(elem, pindex, 8, xlen, hint)
+            return _c_search_hkey_i64(elem, pindex + start * 8, 8, xlen, hint)
```
klaussfreire: `_c_search_hkey_X` methods only exist in cythonized code. This method looks like it shouldn't be called from pure-python ever, so you can make it `cfunc` instead of `ccall`. The calling convention will be faster, and it won't be visible from python, so people cannot shoot themselves in the foot.

sharedbuffers/mapped_struct.py
```diff
@@ -1110,24 +1109,24 @@ def __contains__(self, elem):
...
 
+        lo = 0
+        hi = len(self.objlist)
+
         if cython.compiled:
             dcode, _, _, offset, _ = self.objlist._metadata()
             if dcode in ('q', 'I', 'i', 'd'):
-                i1 = self._contains(elem, dcode, offset)
-                return self.objlist._getitem(i1) == elem
+                lo = self._search_key(elem, dcode, offset, lo, hi, hi // 2)
```
klaussfreire: If we're sorting by stable hash, keys are hashes. As such, you can compute a better hint, by using the same hinting logic StringIdMappers use.

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
llogiudice: But this code is only called when the underlying proxied list is numeric, so the keys aren't the stable hashes. For those cases, there's the basic binary search implemented below.

sharedbuffers/mapped_struct.py
llogiudice: Good idea.

sharedbuffers/mapped_struct.py
klaussfreire: Oh, yes. Sorry, brainfart.

sharedbuffers/mapped_struct.py
```diff
@@ -2014,7 +2024,8 @@ def _subset(self, seq, strict_subset):
                         else:
                             # pfset[j] < val, skip as much as we can
                             while j < seqlen:
-                                if _stable_hash(pfset.objlist._getitem(j)) >= h1:
+                                if _stable_hash(pfset.objlist.__getitem(
+                                    j, dcode2, objlen2, itemsize2, offset2, _struct2, None)) >= h1:
```
klaussfreire: Check indentation

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
```diff
@@ -626,6 +626,10 @@ def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0,
                 buf[offs+1:offs+8] = '\xff\xff\xff\xff\xff\xff\xff'
                 buf[offs+8:offs+12] = _struct_l_Q.pack(objlen)
                 offs += 12
+
+            if sort_fn is not None:
+                obj = sort_fn(obj, False)
```
klaussfreire: Don't forget about the refactoring we talked about re this sort_fn stuff

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
```diff
@@ -472,6 +482,13 @@ def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0):
                 return mapped_tuple.pack_into(tup, buf, offs, idmap, implicit_offs)
         else:
             # Same representation as a tuple of items, only backed in-memory by a frozenset
+            if all_float:
```
klaussfreire, line 490: Probably add to the comment above that it's sorted by stable_hash/value

sharedbuffers/mapped_struct.py
```diff
@@ -530,7 +547,7 @@ class mapped_tuple(tuple):
         offs = cython.Py_ssize_t, implicit_offs = cython.Py_ssize_t, val_offs = cython.Py_ssize_t,
         i = cython.Py_ssize_t, iminval = cython.longlong, imaxval = cython.longlong)
     def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0,
-            array = array.array):
+            int = int, type = type, min = min, max = max, array = array.array, id = id):
```
klaussfreire: Don't do that for int/type/etc in Cython, it hurts Cython optimizations of those builtins

sharedbuffers/mapped_struct.py
```diff
@@ -1719,6 +1737,374 @@ def __str__(self):
...
+        else:
+            raise NotImplementedError("Unsupported data type for fast lookup: %s" % chr(dcode))
+
+    @cython.locals(lo=cython.Py_ssize_t, hi=cython.Py_ssize_t, step=cython.Py_ssize_t,
+        dcode=cython.char, objlen=cython.longlong, itemsize=cython.uchar,
+        offset=cython.Py_ssize_t, h1=cython.ulonglong, h2=cython.ulonglong)
+    def __contains__(self, elem):
+        if self.objlist is None:
+            try:
+                eint = int(elem)
```
klaussfreire: You probably want to declare `eint` as a `ulonglong` (and trap OverflowError as you do with ValueError and TypeError)

sharedbuffers/mapped_struct.py
```diff
@@ -1719,6 +1737,374 @@ def __str__(self):
...
+                if eint != elem or eint >= 128 or eint < 0:
+                    return False
+                return (eint < 64 and (self.bitrep_lo & (1 << eint)) != 0) or (
+                    (eint < 128 and (self.bitrep_hi & (1 << (eint - 64))) != 0))
+            except (ValueError, TypeError):
```
klaussfreire: You should only catch errors in the `eint` cast

sharedbuffers/mapped_struct.py
```diff
@@ -1719,6 +1737,374 @@ def __str__(self):
...
+        hi = len(self.objlist)
+        dcode, objlen, itemsize, offset, _struct = self.objlist._metadata()
+
+        if cython.compiled:
+            if dcode in ('q', 'I', 'i', 'd'):
+                return self._search_key(elem, dcode, offset, lo, hi, hi // 2, True) < hi
+
+        h2 = _stable_hash(elem)
+        while lo < hi:
+            step = (lo + hi) >> 1
```
klaussfreire: step -> mid

And it's:

``` python
mid = lo + ((hi - lo) >> 1)
```

sharedbuffers/mapped_struct.py
```diff
@@ -1719,6 +1737,374 @@ def __str__(self):
...
+            val = self.objlist._c_getitem(step, dcode, objlen, itemsize, offset, _struct, None)
+            h1 = _stable_hash(val)
+            if h1 == h2 and val == elem:
+                return True
```
klaussfreire: If you've found the exact hash, you should stop the binary search and do a bidirectional sequential search for the value. The binary search has outlived it's purpose at this point, all items of same hash should be consecutive. You might be looking at the only element of that hash, or at the middle point of a run of elements of the same hash. So you look sequentially in both directions for the right value, but in any case the binary search is done.

sharedbuffers/mapped_struct.py
```diff
@@ -1719,6 +1737,374 @@ def __str__(self):
...
+            elif h1 < h2:
+                lo = step + 1
+            else:
+                hi = step
+        return False
+
+    @cython.locals(i=cython.Py_ssize_t, val=cython.size_t, xlen=cython.Py_ssize_t,
```
klaussfreire: `val` and `xlen` are unused

sharedbuffers/mapped_struct.py
```diff
@@ -1719,6 +1737,374 @@ def __str__(self):
...
+            i = 0
+
+            while bitrep != 0:
+                if (bitrep & 1) != 0:
+                    yield i
+                bitrep >>= 1
+                i += 1
+
+            bitrep = self.bitrep_hi
+            while bitrep != 0:
```
klaussfreire: You should reset `i`, the above while loop might return early so it may be that `i < 56`. Or, do the above.

sharedbuffers/mapped_struct.py
```diff
@@ -1719,6 +1737,374 @@ def __str__(self):
...
+            for i in xrange(len(self)):
+                yield self.objlist._c_getitem(i, dcode, objlen, itemsize, offset, _struct, None)
+        else:
+            bitrep = self.bitrep_lo
+            i = 0
+
+            while bitrep != 0:
+                if (bitrep & 1) != 0:
+                    yield i
+                bitrep >>= 1
```
klaussfreire: I think it's best if you use an unsigned `i` as induction variable and compute a mask, rather than shift bitrep. Shifting bitrep makes this loop unparallelizable, but if you just compute the mask and test against it, the compiler might unroll/pipeline the loop since it has fewer data dependencies:

``` python
for val in xrange(56):
    if self.bitrep_lo & (1 << val):
        yield val
    elif not self.bitrep_lo >> val:
        break
```

sharedbuffers/mapped_struct.py
```diff
@@ -1719,6 +1737,374 @@ def __str__(self):
...
+    @cython.ccall
+    @cython.locals(i=cython.Py_ssize_t, xlen=cython.Py_ssize_t,
+        pfset='proxied_frozenset', bitrep=cython.size_t, dcode=cython.char,
+        objlen=cython.longlong, itemsize=cython.uchar, offset=cython.Py_ssize_t)
+    def _frozenset_eq(self, x):
+        if isinstance(x, proxied_frozenset):
+            pfset = cython.cast(proxied_frozenset, x)
+            return self.objlist == pfset.objlist and (
+                self.bitrep_lo == pfset.bitrep_lo and
+                self.bitrep_hi == pfset.bitrep_hi)
```
klaussfreire: You should compare the bitmaps first, since that's faster.

Still, make sure to test the case of `frozenset([2L]) == frozenset([2])` with both being proxied frozensets. As we've seen, both could end up using a different representation and break this test.

sharedbuffers/mapped_struct.py
```diff
@@ -1719,6 +1737,374 @@ def __str__(self):
...
+                    return False
+                bitrep >>= 1
+                i += 1
+
+            bitrep = self.bitrep_hi
+            while bitrep != 0:
+                if (bitrep & 1) != 0 and i not in x:
+                    return False
+                bitrep >>= 1
+                i += 1
```
klaussfreire: Same thing about bitmaps here

sharedbuffers/mapped_struct.py
```diff
@@ -1719,6 +1737,374 @@ def __str__(self):
...
+            if pfset.objlist is None:
+                if self.objlist is None:
+                    # Both proxies use compressed representation.
+                    return (self.bitrep_lo & pfset.bitrep_lo) == self.bitrep_lo and (
+                        self.bitrep_hi & pfset.bitrep_hi) == self.bitrep_hi and (
+                        not strict_subset or self.bitlen < pfset.bitlen)
+                else:
+                    dcode, objlen, itemsize, offset, _struct = self.objlist._metadata()
+                    for i in xrange(xlen):
+                        if self.objlist._c_getitem(i, dcode, objlen, itemsize, offset, _struct, None) not in pfset:
```
klaussfreire: You already know pfset is a bitmap. You should extract the bitmap check into a cfunc method you can call in an optimized fashion, and avoid the generic membership test for speed.

Also, if objlist hasn't a numeric typecode, it has non-numeric objects, which means it can never be a subset of a bitmap frozenset, so you could do that quick check. In essence, do a switch on the typecode and summarily return "False" for bitmap-incompatible typecodes.

sharedbuffers/mapped_struct.py
```diff
@@ -1719,6 +1737,374 @@ def __str__(self):
...
+                        return False
+                    bitrep >>= 1
+                    i += 1
+
+                bitrep = self.bitrep_hi
+                while bitrep != 0:
+                    if (bitrep & 1) != 0 and i not in pfset:
+                        return False
+                    bitrep >>= 1
+                    i += 1
```
klaussfreire: Same as above for this case.

sharedbuffers/mapped_struct.py
```diff
@@ -1719,6 +1737,374 @@ def __str__(self):
...
+            if xlen > seqlen:
+                return False
+
+            dcode, objlen, itemsize, offset, _struct = self.objlist._metadata()
+            if dcode in ('q', 'I', 'i', 'd') and cython.compiled:
+                # fast path, use the search_hkey variants
+                dcode2, objlen2, itemsize2, offset2, _struct2 = pfset.objlist._metadata()
+                for i in xrange(xlen):
+                    val = self.objlist._c_getitem(i, dcode, objlen, itemsize, offset, _struct, None)
+                    j = pfset._search_key(val, dcode, offset, j, seqlen - j, j)
```
klaussfreire: Check that search_hkey C functions now have the ability to do that check for you. You should use it, since that's simpler and faster, since it avoids the need for the check below. You can just check `j < objlen2`

sharedbuffers/mapped_struct.py
```diff
@@ -1719,6 +1737,374 @@ def __str__(self):
...
+                    if pfset.objlist._c_getitem(j, dcode2, objlen2, itemsize2, offset2, _struct2, None) != val:
```
klaussfreire: If search_hkey fails to find a key, it may retun an index beyond the end. Then getitem would raise an IndexError. Beware of that, you have to verify the returned index.

sharedbuffers/mapped_struct.py
```diff
@@ -1719,6 +1737,374 @@ def __str__(self):
...
+                        if h1 == h2 and val == val2:
+                            break
+                        elif h1 < h2 or j >= seqlen:
+                            return False
+                        else:
+                            # pfset[j] < val, skip as much as we can
+                            while j < seqlen:
+                                if _stable_hash(pfset.objlist._c_getitem(j, dcode2, objlen2, itemsize2, offset2, _struct2, None)) >= h1:
+                                    break
+                                j += 1
```
klaussfreire, line 2192: You should skip exponentially faster until you've skipped too much, then backtrack. As search_hkey does.

sharedbuffers/mapped_struct.py
```diff
@@ -2122,6 +2508,7 @@ class mapped_object(object):
         proxied_ndarray: 'n',
         proxied_buffer: 'r',
         proxied_dict: 'M',
+        proxied_frozenset: 'O',
```
klaussfreire: O looks like Object. Is lowercase `z` taken?

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
llogiudice: I'm pretty sure this is the result of a merge, since I don't recall modifying these lines (I have no reason to).

sharedbuffers/mapped_struct.py
llogiudice: I thought cython was smart enough to figure that out. Oh well.

klaussfreire: `int` tries to be too smart to be provably `int`. Try it:

``` pycon
>>> type(int('34289739428734928374923874932847'))
<type 'long'>
```

sharedbuffers/mapped_struct.py
llogiudice: Eh, I don't know. I think unrolling is perfectly feasible the way it's done, since the compiler can see that ```i``` is monotonically increased on each operation.

sharedbuffers/mapped_struct.py
llogiudice: I went for 'O' as in 'ordered', but I guess it can be a bit confusing. Lowercase 'z' doesn't appear to be taken.

sharedbuffers/mapped_struct.py
```diff
@@ -1831,19 +1832,24 @@ def _search_key(self, elem, dcode, offset, start, xlen, hint, equal=False):
         else:
             raise NotImplementedError("Unsupported data type for fast lookup: %s" % chr(dcode))
 
-    @cython.locals(lo=cython.Py_ssize_t, hi=cython.Py_ssize_t, step=cython.Py_ssize_t,
+    @cython.locals(eint=cython.ulonglong)
+    def _contains_compressed(self, elem):
```
klaussfreire, line 1850: You want this to be a cfunc so it's a fast call, and you want in that case `elem` to be a native int already too so when you're looping over native ints you don't have to build a python int just to call this method

sharedbuffers/mapped_struct.py
```diff
@@ -1855,41 +1861,58 @@ def __contains__(self, elem):
...
+                    val = self.objlist._c_getitem(mid, dcode, objlen, itemsize, offset, _struct, None)
+                    h1 = _stable_hash(val)
+                    if h1 != h2:
+                        break
+                    elif val == elem:
+                        return True
+                    mid -= 1
+
+                mid = lo + ((hi - lo) >> 1)
+                hi = len(self.objlist)
```
klaussfreire: You don't need this. You already know there's no match behond `hi` whatever value it has.

sharedbuffers/mapped_struct.py
```diff
@@ -1855,41 +1861,58 @@ def __contains__(self, elem):
...
             dcode, objlen, itemsize, offset, _struct = self.objlist._metadata()
             for i in xrange(len(self)):
                 yield self.objlist._c_getitem(i, dcode, objlen, itemsize, offset, _struct, None)
         else:
-            bitrep = self.bitrep_lo
-            i = 0
-
-            while bitrep != 0:
-                if (bitrep & 1) != 0:
+            for i in xrange(64):
```
klaussfreire, line 1937: Cheap optimization:

``` python
if self.bitrep_lo:
    for i in xrange(64):
        ...
if self.bitrep_hi:
    for i in xrange(64):
        ...
```

sharedbuffers/mapped_struct.py
```diff
@@ -1969,25 +1989,28 @@ def _subset(self, seq, strict_subset):
                         not strict_subset or self.bitlen < pfset.bitlen)
                 else:
                     dcode, objlen, itemsize, offset, _struct = self.objlist._metadata()
+                    if dcode not in ('q', 'I', 'i', 'd'):
```
klaussfreire: `I` and `i` are numeric... or did you mean to say `t` and `T`?

sharedbuffers/mapped_struct.py
```diff
@@ -2003,9 +2026,10 @@ def _subset(self, seq, strict_subset):
                 dcode2, objlen2, itemsize2, offset2, _struct2 = pfset.objlist._metadata()
                 for i in xrange(xlen):
                     val = self.objlist._c_getitem(i, dcode, objlen, itemsize, offset, _struct, None)
-                    j = pfset._search_key(val, dcode, offset, j, seqlen - j, j)
-                    if pfset.objlist._c_getitem(j, dcode2, objlen2, itemsize2, offset2, _struct2, None) != val:
+                    tmp_idx = pfset._search_key(val, dcode, offset, j, seqlen - j, j, True)
+                    if not tmp_idx < seqlen - j:
```
klaussfreire: I prefer `tmp_idx >= seqlen - j`. Other code uses that IIRC.

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
llogiudice: Hm. I was thinking of the most pathological case possible, in which all elements of the list have the same hash value, but are all different. In that case, I would have to look both to the left and the right ... right ?

sharedbuffers/mapped_struct.py
llogiudice: ```dcode in ('t', 'T')``` is semantically equivalent to ```dcode not in ('q', 'I', 'i', 'd')```
Since I used the latter tuple below, I figured it would be more clear this way.

sharedbuffers/mapped_struct.py
llogiudice: Sure.

sharedbuffers/mapped_struct.py
```diff
@@ -403,24 +403,34 @@ def get_wrapped_key(obj):
...
+            all_int = all_intlong = all_float = 0
             obj_dtype = obj.dtype
             if obj_dtype.isbuiltin:
                 dtype = obj_dtype.char
                 cdtype = cython.cast('const char*', dtype)[0]
-                if cdtype in ('L', 'l', 'I', 'i', 'H', 'h', 'B', 'b'):
+                if cdtype in ('l', 'I', 'i', 'H', 'h', 'B', 'b'):
                     all_int = 1
+                elif cdtype == 'L':
+                    all_intlong = 1
```
tzulberti-jampp, line 423: I am not sure this are equivalent. In the `else` on line 422, `all_int = 1` implies that `all_intlog =1`. But in this case, you might have `all_int = 1` and `all_intlong = 0`.

sharedbuffers/mapped_struct.py
tzulberti-jampp: Ping about this. Have you checked if those changes are from another branch or what?

sharedbuffers/mapped_struct.py
```diff
@@ -1504,7 +1523,7 @@ def _getitem(self, index):
         lpindex = "const long *",
         ipindex = "const int *",
         dataoffs = cython.Py_ssize_t, itemsize = cython.uchar)
-    def __getitem(self, index, dcode, objlen, itemsize, dataoffs, _struct, proxy_into):
+    def _c_getitem(self, index, dcode, objlen, itemsize, dataoffs, _struct, proxy_into):
```
tzulberti-jampp, line 1530: Why the rename of this function?

sharedbuffers/mapped_struct.py
```diff
@@ -1719,6 +1738,403 @@ def __str__(self):
...
+
+    def symmetric_difference(self, *seqs):
+        return frozenset(self).symmetric_difference(*seqs)
+
+    @cython.ccall
+    @cython.locals(i=cython.Py_ssize_t, xlen=cython.Py_ssize_t,
+        pfset='proxied_frozenset', dcode=cython.char,
+        objlen=cython.longlong, itemsize=cython.uchar, offset=cython.Py_ssize_t)
+    def _frozenset_eq(self, x):
+        if isinstance(x, proxied_frozenset):
```
tzulberti-jampp, line 2066: Shouldn't we use `if type(x) is proxied_frozenset` to make it faster?

sharedbuffers/mapped_struct.py
```diff
@@ -1719,6 +1738,403 @@ def __str__(self):
...
+        return self.symmetric_difference(seq)
+
+    def __rxor__(self, seq):
+        return frozenset(seq).symmetric_difference(self)
+
+    def __repr__(self):
+        return "proxied_frozenset(%s)" % str(self)
+
+    def __str__(self):
+        return "[%s]" % ",".join(str(x) for x in self)
```
tzulberti-jampp: It is wired that you are using the `[]`, since those are used for list. I would put something like `set([...])` that is the value returned by the pure python sets

sharedbuffers/mapped_struct.py
```diff
@@ -1719,6 +1738,403 @@ def __str__(self):
...
+                        else:
+                            # pfset[j] < val, skip as much as we can
+                            while j < seqlen:
+                                if _stable_hash(pfset.objlist._c_getitem(j, dcode2, objlen2, itemsize2, offset2, _struct2, None)) >= h1:
+                                    break
+                                j += 1
+                            if j == seqlen:
+                                return False
+            return not strict_subset or xlen < seqlen
+        elif xlen > len(seq):
```
tzulberti-jampp: Why do you know that `seq` has the `len` attribute? For example, on Python 2, you could do the following:
```
>>> set([1,2,3]) < 3
False
```

sharedbuffers/mapped_struct.py
```diff
@@ -1719,6 +1738,403 @@ def __str__(self):
...
+                        j += 1
+
+                        if h1 == h2 and val == val2:
+                            break
+                        elif h1 < h2 or j >= seqlen:
+                            return False
+                        else:
+                            # pfset[j] < val, skip as much as we can
+                            while j < seqlen:
+                                if _stable_hash(pfset.objlist._c_getitem(j, dcode2, objlen2, itemsize2, offset2, _struct2, None)) >= h1:
```
tzulberti-jampp: I think this line might be too long... maybe we can split it into 2?

sharedbuffers/mapped_struct.py
llogiudice: Good catch.

sharedbuffers/mapped_struct.py
llogiudice: Because the double underscores would limit things like inheritance, so I went with the common convention of naming `_c_something`.

sharedbuffers/mapped_struct.py
llogiudice: we could, but that wouldn't work for subclasses, I think.

sharedbuffers/mapped_struct.py
llogiudice: You mean just like the `repr` method ?

sharedbuffers/mapped_struct.py
llogiudice: Ah, you're right. I'll fix that.

sharedbuffers/mapped_struct.py
llogiudice: Sure.

sharedbuffers/mapped_struct.py
llogiudice: It was apparently removed at some point (older branches have it), so I guess I'll just take it out.

sharedbuffers/mapped_struct.py
tzulberti-jampp: At least in Python 2, for sets the repr and str returns the same things:

``` 
>>> s = set([1,2,3])
>>> print '%r'% s
set([1, 2, 3])
>>> print '%s' % s
set([1, 2, 3])
```

llogiudice: That's good to know. I'll change `__str__`, then.

sharedbuffers/mapped_struct.py
```diff
@@ -2060,7 +2068,7 @@ def _subset(self, seq, strict_subset):
                             if j == seqlen:
                                 return False
             return not strict_subset or xlen < seqlen
-        elif not hasattr(seq, '__len__') or xlen > len(seq):
+        elif proxied_frozenset._safe_len_gt(xlen, seq):
             return False
         elif self.objlist is None:
```
``` python
else:
    try:
        seqlen = len(seq)
    except (AttributeError, TypeError):
        return False
    else:
        if xlen > seqlen:
            return False
        elif self.objlist is None:
            ...
```

klaussfreire (changes requested):

klaussfreire (approved):

klaussfreire (changes requested):
Forgot you still need to merge fused_types

tests/mapped_struct.py
```diff
@@ -1570,6 +1558,91 @@ def testIndexedCollectionEqual(self):
...
+        self.assertTrue(big > small)
+        self.assertTrue(big >= small)
+        self.assertFalse(big > big)
+
+        big = self.pack([1., 1.5, 2., 2.5, 3.])
+        self.assertTrue(small < big)
+        self.assertTrue(big > small)
+
+    def testCompressed(self):
+        small = self.pack([1, 2])
```
gefarion, line 1619: Add big numbers to force usage of both bitmaps.

sharedbuffers/mapped_struct.py
```diff
@@ -1719,6 +1742,420 @@ def __str__(self):
...
+                        break
+
+            if self.bitrep_hi:
+                for i in xrange(64):
+                    if self.bitrep_hi & (1 << i):
+                        yield i + 64
+                    elif not self.bitrep_hi >> i:
+                        break
+
+    def union(self, *seqs):
```
gefarion, line 2029: @klaussfreire It is ok to maintain these operations slow?

tests/mapped_struct.py
```diff
@@ -1570,6 +1558,91 @@ def testIndexedCollectionEqual(self):
...
+        self.assertEqual(c.intersection([3., 5., 9.]), frozenset([3., 5.]))
+        self.assertEqual(c.difference([2., 4., 8.]), frozenset([1., 3., 5.]))
+        self.assertEqual(c.symmetric_difference([1., 8.]), frozenset([2., 3., 4., 5., 8.]))
+
+    def testSubsetSuperset(self):
+        small = self.pack([1., 2.])
+        big = self.pack([1., 2., 3., 4.])
+        self.assertTrue(small < big)
+        self.assertTrue(small <= big)
+        self.assertFalse(small < small)
```
gefarion, line 1606: I would add `small <= small`

tests/mapped_struct.py
```diff
@@ -1570,6 +1558,91 @@ def testIndexedCollectionEqual(self):
...
+        self.assertEqual(c, frozenset(c))
+
+    def testSetOps(self):
+        c = self.pack([1., 2., 3., 4., 5.])
+        self.assertEqual(c.union([3., 6.]), frozenset([1., 2., 3., 4., 5., 6.]))
+        self.assertEqual(c.intersection([3., 5., 9.]), frozenset([3., 5.]))
+        self.assertEqual(c.difference([2., 4., 8.]), frozenset([1., 3., 5.]))
+        self.assertEqual(c.symmetric_difference([1., 8.]), frozenset([2., 3., 4., 5., 8.]))
+
+    def testSubsetSuperset(self):
```
gefarion, line 1601: I would add test cases with an empty set in all operation tests.

tests/mapped_struct.py
```diff
@@ -1570,6 +1558,91 @@ def testIndexedCollectionEqual(self):
...
+
+    def testContains(self):
+        values = [float(x) for x in range(5)]
+        c = self.pack(values)
+        for val in values:
+            self.assertIn(val, c)
+        for val in values:
+            self.assertNotIn(-val - 1, c)
+
+    def testCmpSets(self):
```
gefarion, line 1573: Add a test case comparing with an invalid object and catch the exception.

tests/mapped_struct.py
llogiudice: This particular test does check for correctness when both operands use bitmaps.

tests/mapped_struct.py
llogiudice: Sure.

tests/mapped_struct.py
llogiudice: Python 2 doesn't raise an exception when comparing against non-sets, it just returns an arbitrary, but consistent value.

tests/mapped_struct.py
llogiudice: OK.

tests/mapped_struct.py
gefarion: Yes, if I am right, you have two bitmaps in each compressed frozenset. I suggest adding numbers that force the usage these both bitmaps.

llogiudice: That's the thing. For numbers between 0 and 120, the proxied frozensets use bitmaps. So that test is already doing the check you requested.

tests/mapped_struct.py
gefarion: I would add these for the comparison operators too.

tests/mapped_struct.py
klaussfreire: Not really. The higher portion of the bitmap is 0 here and will be skipped by the optimization you added later. Run coverage if you prefer. You should add a set with higher (yet bitmap-able) values so that the higher portion of the bitmap gets used.

sharedbuffers/mapped_struct.py
klaussfreire: On the one side, most of these operations have to create a new set, so they're screwed in that sense.

On the other side, it's possible we can implement optimized construction of that new set at least. We will use them, so we want them as fast as practical.

For instance, in the case of bitmap sets on both sides, most operations can be extremely efficient. Most other cases are tougher though.

klaussfreire: At the very least, a fast-call version of `frozenset(self)` that avoids having to iterate self through the pure-python interface could be added.

sharedbuffers/mapped_struct.py
```diff
@@ -1719,6 +1742,420 @@ def __str__(self):
...
+
+            return True
+
+        dcode, objlen, itemsize, offset, _struct = self.objlist._metadata()
+        for i in xrange(xlen):
+            if self.objlist._c_getitem(i, dcode, objlen, itemsize, offset, _struct, None) not in x:
+                return False
+        return True
+
+    def __eq__(self, seq):
```
klaussfreire, line 2098: You're missing `__ne__`

klaussfreire (changes requested):

tests/mapped_struct.py
```diff
@@ -1575,6 +1575,11 @@ def testCmpSets(self):
         self.assertEqual(c, self.pack([5., 4., 3., 2., 1.]))
         self.assertEqual(c, set(c))
         self.assertEqual(c, frozenset(c))
+        self.assertNotEqual(c, set())
+        self.assertNotEqual(c, frozenset())
+        c = self.pack([])
+        self.assertEqual(c, set())
+        self.assertEqual(c, frozenset())
```
klaussfreire, line 1582:
```
self.assertNotIn(3, c)
```

tests/mapped_struct.py
llogiudice: Oh, you meant usage of both `bitmap_lo` and `bitmap_hi`. I'm dumb.

sharedbuffers/mapped_struct.py
```diff
@@ -1930,52 +1930,75 @@ def __iter__(self):
...
-            pfset = cython.cast(proxied_frozenset, seqs[0])
+    @cython.ccall
+    @cython.locals(pfset='proxied_frozenset')
+    def _union_2(self, seq):
+        if self.objlist is None and type(seq) is proxied_frozenset:
+            pfset = cython.cast(proxied_frozenset, seq)
             if pfset.objlist is None:
                 return proxied_frozenset(
                     None, self.bitrep_lo | pfset.bitrep_lo, self.bitrep_hi | pfset.bitrep_hi)
+        return frozenset(self).union(seq)
```
klaussfreire: You'll be iterating `self` twice at least. Once to build the left frozenset, and then to compute the union and build the resulting frozenset.

That should be akin to:

``` python
rv = set()
self._add_to(rv)
for elem in seq:
    elem._add_to(rv)
return frozenset(rv)
```

In performance. Why don't you test it?

Same in all other operations. You're paying the construction cost twice, once for the temporary, and once for the result. So it may be OK to go through a regular set first and optimize the operation.

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
llogiudice: Creating a set with fast iteration and performing inplace updates on it does seem to improve performance. What doesn't have as much impact is using the `merge` interface with integers (only 2 to 5% improvement). I would like to know your opinion, but I don't think it's worth it.

klaussfreire: If it isn't worth it, it isn't worth it. We can revisit it later in any case.

Paste the timeit script and timings if you can though.

sharedbuffers/mapped_struct.py
```diff
@@ -1918,27 +1918,53 @@ def __iter__(self):
         else:
             if self.bitrep_lo:
                 for i in xrange(64):
-                    if self.bitrep_lo & (1 << i):
+                    if self.bitrep_lo & (cython.cast(cython.size_t, 1) << i):
```
klaussfreire: Use the same type of bitrep_lo (ulonglong)

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
llogiudice: Here's the script I used:
```python
from sharedbuffers import mapped_struct
from functools import partial
import timeit

def make_obj(start, end):
    base = [x * 2 for x in xrange(start, end)] + [2 ** 30]
    buf = bytearray(1024 * 16)
    mapped_struct.proxied_frozenset.pack_into(frozenset(base), buf, 0)
    return mapped_struct.proxied_frozenset.unpack_from(buf, 0)

if __name__ == '__main__':
    obj1 = make_obj(0, 1000)
    obj2 = make_obj(500, 1500)
    print "union: ", timeit.timeit(partial(obj1.union, obj2))
```

And the timings:
- Naive: 126.80619231505
- Fast conversion: 110.266254187
- Merge: 105.635071511146
- CPython frozensets: 97.6680510044

sharedbuffers/mapped_struct.py
```diff
@@ -2171,7 +2708,8 @@ def p(s):
         'v' : (mapped_datetime.pack_into, mapped_datetime.unpack_from, mapped_datetime),
         'V' : (mapped_date.pack_into, mapped_date.unpack_from, mapped_date),
         'F' : (mapped_decimal.pack_into, mapped_decimal.unpack_from, mapped_decimal),
-        'M' : (proxied_dict.pack_into, proxied_dict.unpack_from, proxied_dict)
+        'M' : (proxied_dict.pack_into, proxied_dict.unpack_from, proxied_dict),
+        'z' : (proxied_frozenset.pack_into, proxied_frozenset.unpack_from, proxied_frozenset)
```
tzulberti-jampp: Nit: I would add a comma at the end of the line so the next time that he have to add a value then there is no diff to
an unrelated line

sharedbuffers/mapped_struct.py
```diff
@@ -4265,251 +4590,159 @@ def sorted_contains(a, hkey):
...
         pend1 = pindex1 + stride0 * length1
         pend2 = pindex2 + stride0 * length2
         pdestend = pdest + stride0 * destlength
         pdeststart = pdest
         while pindex1 < pend1 and pindex2 < pend2 and pdest < pdestend:
-            ref = cython.cast(cython.p_ulonglong, pindex2)[0]
-            while pindex1 < pend1 and cython.cast(cython.p_ulonglong, pindex1)[0] <= ref and pdest < pdestend:
-                cython.cast(cython.p_ulonglong, pdest)[0] = cython.cast(cython.p_ulonglong, pindex1)[0]
-                cython.cast(cython.p_ulonglong, pdest)[1] = cython.cast(cython.p_ulonglong, pindex1)[1]
+            ref = cython.cast('numeric_A *', pindex2)[0]
```
tzulberti-jampp, line 4682: You are overriding the `ref` parameter?

tzulberti-jampp (approved):
Some nitpicking comments , but you have my +1

sharedbuffers/mapped_struct.py
llogiudice: Well, remember that `ref` is a fused-type parameter, and thus, it's only used to have cython deduce the correct type. It's not useful in that sense. But, since I need a variable of that very type, I might as well use it ;)

sharedbuffers/mapped_struct.py
llogiudice: Sure.

sharedbuffers/mapped_struct.py
tzulberti-jampp: ok, I miss that

sharedbuffers/mapped_struct.py
klaussfreire: I'd say "Fast conversion" is worth it, though going any further than that probably doesn't gain enough to offset the increased complexity.

llogiudice: Agreed. Btw, 'fast conversion' refers to what we're currently using, in which we build a set via C-style iteration instead of generators.

sharedbuffers/mapped_struct.py
```diff
@@ -1718,6 +1741,519 @@ def __str__(self):
...
+        if not seq:
+            return self
+        elif self.objlist is None and type(seq) is proxied_frozenset:
+            pfset = cython.cast(proxied_frozenset, seq)
+            if pfset.objlist is None:
+                return proxied_frozenset(
+                    None, self.bitrep_lo | pfset.bitrep_lo, self.bitrep_hi | pfset.bitrep_hi)
+
+        rv = self._add_to(set())
+        rv.update(seq)
```
klaussfreire: I think it *should* be faster

``` python
return frozenset(self._add_to(set(seq)))
```

Because set copying is optimized in CPython (it just copies the hash table).

sharedbuffers/mapped_struct.py
```diff
@@ -1718,6 +1741,519 @@ def __str__(self):
...
+        if self.objlist is None and type(seq) is proxied_frozenset:
+            pfset = cython.cast(proxied_frozenset, seq)
+            if pfset.objlist is None:
+                return proxied_frozenset(
+                    None, self.bitrep_lo & pfset.bitrep_lo, self.bitrep_hi & pfset.bitrep_hi)
+        elif not seq:
+            return proxied_frozenset(None, 0, 0)
+
+        rv = self._add_to(set())
+        rv.intersection_update(seq)
```
klaussfreire: You should check the which is smallest and iterate the smallest one, checking for membership against the big one.

Basically the following, except with optimized iteration/membership like in add_to:

``` python
if len(self) < len(seq):
    rv = set()
    for elem in self:
        if elem in seq:
            rv.add(seq)
else:
    for elem in seq:
        if elem in self:
            rv.add(elem)
    pass
return frozenset(rv)
```

sharedbuffers/mapped_struct.py
```diff
@@ -1718,6 +1741,519 @@ def __str__(self):
...
+        if not seq:
+            return self
+        elif self.objlist is None and type(seq) is proxied_frozenset:
+            pfset = cython.cast(proxied_frozenset, seq)
+            if pfset.objlist is None:
+                return proxied_frozenset(
+                    None, self.bitrep_lo & ~pfset.bitrep_lo, self.bitrep_hi & ~pfset.bitrep_hi)
+
+        rv = self._add_to(set())
+        rv.difference_update(seq)
```
klaussfreire, line 2011: Same here, iterate the smallest, check membership against the biggest.

sharedbuffers/mapped_struct.py
```diff
@@ -1718,6 +1741,519 @@ def __str__(self):
...
+            return self
+        elif self.objlist is None and type(seq) is proxied_frozenset:
+            pfset = cython.cast(proxied_frozenset, seq)
+            if pfset.objlist is None:
+                return proxied_frozenset(
+                    None, self.bitrep_lo ^ pfset.bitrep_lo, self.bitrep_hi ^ pfset.bitrep_hi)
+
+        rv = self._add_to(set())
+        rv.symmetric_difference_update(seq)
+        return frozenset(rv)
```
klaussfreire, line 2027: I don't particularly care about this one, but you can do it as: `(self - seq) | (seq - self)` I think

sharedbuffers/mapped_struct.py
```diff
@@ -1718,6 +1741,519 @@ def __str__(self):
...
+
+    def union(self, *seqs):
+        if not seqs:
+            return self
+        elif len(seqs) == 1:
+            return self._union_2(seqs[0])
+        return frozenset(self).union(*seqs)
```
klaussfreire: More like

``` python
else:
    return frozenset(self._add_to(set())).union(*seqs)
```

sharedbuffers/mapped_struct.py
```diff
@@ -1718,6 +1741,519 @@ def __str__(self):
...
+
+    def intersection(self, *seqs):
+        if not seqs:
+            return self
+        elif len(seqs) == 1:
+            return self._intersect_2(seqs[0])
+        return frozenset(self).intersection(*seqs)
```
klaussfreire: Again, you ought to implement incremental intersection with intersect_2, or you won't reap the benefits of any optimization in intersect_2.

Same with all the other methods.

Many of them don't even require using the _2 version as a base. Difference can be implemented just as difference_2 only you check for membership against multiple sequences rather than just one.

Try to have good implementations for all the operations.

klaussfreire (changes requested):
When you're done with the changes, re-run the timings.

sharedbuffers/mapped_struct.py
llogiudice: I tried this, but there's no gain, because Python's sets already do this: https://github.com/python/cpython/blob/master/Objects/setobject.c#L1256

sharedbuffers/mapped_struct.py
llogiudice: This does seem to generate better code. Thanks.

sharedbuffers/mapped_struct.py
llogiudice: I've done some measuring already, and I focused on making the performance of the 2 argument functions because from what I could tell, once you get into variadic functions, it's a lost battle because of CPython's calling convention.

sharedbuffers/mapped_struct.py
```diff
@@ -1976,8 +1975,27 @@ def _intersect_2(self, seq):
         elif not seq:
             return proxied_frozenset(None, 0, 0)
 
-        rv = self._add_to(set())
-        rv.intersection_update(seq)
+        if len(self) > len(seq):
+            rv = set(seq)
+            typ = type(seq)
```
klaussfreire: Don't assign it to a local, that hurts rather than help. Cython knows how to optimize `type(x) is y` very well.

sharedbuffers/mapped_struct.py
```diff
@@ -1976,8 +1975,27 @@ def _intersect_2(self, seq):
...
+
+            if typ is set:
+                set_seq = cython.cast(set, seq)
```
klaussfreire: cython.cast is weird in pure python, just declare the type of set_seq in locals and leave the cast implicit

klaussfreire (approved):

PR #63, opened 2018-11-22
Fix Schema.pack thread safety

klaussfreire:


sharedbuffers/mapped_struct.py
```diff
@@ -3277,20 +3286,38 @@ def pack_into(self, obj, buf, offs, idmap = None, packer = None, padding = None,
             self.postwrite_hook(obj, buf, baseoffs, offs)
         return offs
 
+    @cython.cfunc
+    def _acquire_pack_buffer(self):
```
llogiudice, line 3290: Does cython make any guarantees regarding the GIL when executing cfunc's ?

klaussfreire: Nope. But this is written to avoid releasing the GIL between reading and writing the attribute.

1. The attribute is a cython attribute, so reading it is natively done, without python involvement
2. No deallocations can occur before writing the attribute (we hold onto a reference to the buffer), so no GC invocations
3. Writing is also natively done

klaussfreire: But yeah, it only works if this is Cythonized, I could clarify that in the changelog

llogiudice (approved):

gefarion (approved):

PR #67, opened 2018-11-27
Fused types

llogiudice:
What do you get when you combine Python, C, and something that resembles C++'s templates ?

NOTE: Cython's rule for fused types are quite different from C++ templates in that they only work on parameters' types (There's no free-style instantiation, for example). If you see spurious variables being declared for no apparent reason, don't panic ! That's just how this kludge is supposed to work.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,6 +31,7 @@ Added
   through proxied_list s containing objects, by allowing proxy
   reuse and masked iteration. The first one can take a proxy to use,
   while the second one will build its own generic proxy.
+- Use fused types to reduce code bloat and support more integer formats.
 
 Changes
 -------
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -4174,6 +4174,10 @@ class MappedArrayProxyBase(_ZipMapBase):
 
 if cython.compiled:
 
+    # We need as many of these definitions as different parameters are
+    # used per-function; otherwise, cython will deduce that they are of
+    # the same type. Sigh ...
+
     numeric_A = cython.fused_type(
         cython.char,
         cython.uchar,
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -6628,7 +6628,8 @@ class StringIdMultiMapper(StringIdMapper):
 
     @cython.locals(
         hkey = cython.ulonglong, startpos = int, nitems = int, bkey = bytes,
-        blen = cython.size_t, pbkey = 'const char *', pybuf = 'Py_buffer')
+        blen = cython.size_t, pbkey = 'const char *', pybuf = 'Py_buffer', indexbuf = 'Py_buffer',
+        stride0 = cython.size_t, stride1 = cython.size_t, pindex = cython.p_char)
     def get_iter(self, key, default = None):
         if not isinstance(key, basestring):
             return
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -4565,6 +4565,50 @@ if cython.compiled:
     #@cython.cfunc
     @cython.locals(
         length1 = cython.size_t, length2 = cython.size_t, destlength = cython.size_t,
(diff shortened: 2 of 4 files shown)


sharedbuffers/mapped_struct.py
```diff
@@ -4688,251 +4421,114 @@ def sorted_contains(a, hkey):
...
+    @cython.returns(cython.size_t)
     @cython.locals(
         length1 = cython.size_t, length2 = cython.size_t, destlength = cython.size_t,
-        stride0 = cython.size_t, ref = cython.ulonglong,
+        stride0 = cython.size_t, ref = numeric_A,
         pindex1 = cython.p_char, pindex2 = cython.p_char, pdest = cython.p_char,
         pend1 = cython.p_char, pend2 = cython.p_char, pdestend = cython.p_char, pdeststart = cython.p_char)
-    #@cython.returns(cython.size_t)
-    def _c_merge_ui64(pindex1, length1, pindex2, length2, pdest, destlength, stride0):
+    def _c_merge_gen(ref, pindex1, length1, pindex2, length2, pdest, destlength, stride0):
```
klaussfreire: I'd rather have it in the end (ref)

klaussfreire (changes requested):

klaussfreire (approved):

sharedbuffers/mapped_struct.py
```diff
@@ -4287,299 +4255,105 @@ def _c_search_hkey_i64(hkey, pindex, stride0, length, hint, check_equal):
...
-        pindex = cython.p_char, skip = cython.size_t, check_equal = cython.bint)
-    @cython.returns(cython.size_t)
-    def _c_search_hkey_ui32(hkey, pindex, stride0, length, hint, check_equal):
-        hi = length
-        lo = 0
+    @cython.locals(hkey = cython.ulonglong, elem = cython.ulonglong,
+        pindex = cython.p_char, stride0 = cython.size_t, length = cython.size_t,
+        hint = cython.size_t, check_equal = cython.bint)
+    def _c_search_hkey_ui64(hkey, pindex, stride0, length, hint, check_equal):
+        elem = 0
```
tzulberti-jampp, line 4287: Why do you need to do this initialization on each function?

sharedbuffers/mapped_struct.py
```diff
@@ -4165,119 +4173,79 @@ def map_file(cls, fileobj, offset = 0, size = None):
...
-                hi = mid
-            else:
-                while mid > lo and cython.cast(cython.p_ulonglong, pindex + stride0 * (mid-1))[0] == hkey:
-                    mid -= 1
-                return mid
-        # Check equality if requested
-        if check_equal and lo < length and cython.cast(cython.p_ulonglong, pindex + stride0 * lo)[0] != hkey:
-            lo = length
-        return lo
+    numeric_B = cython.fused_type(
```
tzulberti-jampp, line 4196: I understand that this happens because of the limitations on Cython when using Fused Types. I would add some comment about that in the code to make it explicit

tzulberti-jampp (commented):
Changelog update?

sharedbuffers/mapped_struct.py
llogiudice: When using a function that uses a fused type, cython requires one of the function parameters to be of the given fused type (This is what I meant when I remarked that cython has no 'free-style' instantiation, unlike C++). Thus the need for a 'dummy' variable for each call.

sharedbuffers/mapped_struct.py
llogiudice: Sure.

CHANGELOG.rst
```diff
@@ -31,6 +31,7 @@ Added
   through proxied_list s containing objects, by allowing proxy
   reuse and masked iteration. The first one can take a proxy to use,
   while the second one will build its own generic proxy.
+- Use fused types to reduce code bloat and support more integer formats.
```
tzulberti-jampp, line 34: I think this is more like a Technical Task than something that was added.

tzulberti-jampp (approved):

PR #70, opened 2018-12-17
Fix access to empty idmappers

klaussfreire:
It would segfault under Cython


tzulberti-jampp (approved):

sharedbuffers/mapped_struct.py
```diff
@@ -5306,9 +5308,13 @@ def get(self, key, default = None):
             return default
         if key < 0 or key > self.dtypemax:
             return default
+
         hkey = key
-        startpos = self._search_hkey(hkey)
         nitems = self.index_elements
+        if nitems == 0:
```
llogiudice, line 5314: Are these checks (i.e: the ones happening inside the `get` methods) necessary ? I'd figure that since the `_search_hkey` methods are already doing that, you'd never get a match inside `get`, making this a bit redundant.

klaussfreire: Depends on each implementation. At least for multi-mappers if the check is done in _search_hkey the response is slightly different (an empty list rather than the default)

llogiudice (approved):

PR #77, opened 2019-05-28
Support atomic writes in proxy properties.

llogiudice:
This PR aims to provide atomic writes of integral types (anything specified by the `numeric_A` fused type) in proxy property types. The basic idea is to use memory fences to synchronize, since any sane architecture should support RW atomicity for these types already. Therefore, the only obstacle would be to provide a 'happens-before' relationship by issuing a CPU barrier before the load and after the store.

While we are at it, I also simplified the `__get__` method of these types because repetition drives me nuts.


sharedbuffers/mapped_struct.py
```diff
@@ -3236,7 +3236,9 @@ def _c_buffer_proxy_get_gen(self, obj, elem):
     @cython.cfunc
     @cython.locals(self = BaseBufferProxyProperty, elem = numeric_A, obj = BufferProxyObject)
     def _c_buffer_proxy_set_gen(self, obj, elem):
-        if obj is not None and not (self.none_bitmap & self.mask):
+        if obj.pybuf.readonly:
+            raise TypeError('cannot set object in read-only buffer')
```
klaussfreire: set attribute

sharedbuffers/mapped_struct.py
```diff
@@ -3177,124 +3177,214 @@ def __set__(self, obj, value):
...
+    def _c_buffer_proxy_get_gen(self, obj, elem):
+        if obj is None:
+            return self
+        elif obj.none_bitmap & self.mask:
+            return None
+        else:
+            assert (obj.offs + self.offs + cython.sizeof(elem)) <= obj.pybuf.len   #lint:ok
+            mfence_full()   # acquire
+            return cython.cast('numeric_A *',
+                cython.cast(cython.p_uchar, obj.pybuf.buf) + obj.offs + self.offs)[0]   #lint:ok
```
klaussfreire: I think we should compute the pointer before the fence, and only do

``` python
pvalue = ...
mfence_full()
return pvalue[0]
```

Otherwise, the compiler/cpu will have to re-read all proxy attributes from memory.

klaussfreire (changes requested):
You have to add support for mapping files as writable as well

sharedbuffers/mapped_struct.py
llogiudice: Good idea.

klaussfreire (approved):

sharedbuffers/mapped_struct.py
```diff
@@ -164,7 +167,8 @@ def _likebuffer(buf):
     """
     Takes a buffer object as parameter and returns a writable object with buffer protocol.
     """
-    if type(buf) is buffer or type(buf) is bytearray or type(buf) is bytes or isinstance(buf, bytes):
+    if type(buf) is buffer or type(buf) is bytearray or type(buf) is bytes or (
```
tzulberti-jampp: Why the parenthis on the last 2 possible values? The opening ( should start when the if starts.

sharedbuffers/mapped_struct.py
```diff
@@ -164,7 +167,8 @@ def _likebuffer(buf):
...
+            isinstance(buf, bytes) or isinstance(buf, ctypes_Array)):
```
tzulberti-jampp: You could also do `isinstance(buf, (bytes, ctypes_Array))`

sharedbuffers/mapped_struct.py
```diff
@@ -4586,14 +4700,14 @@ def getter(self, proxy_into = None, no_idmap = False):
...
-        buf = self.buf
+        buf = self.wr_buf
 
         if proxy_class is not None:
             proxy_class_new = functools.partial(proxy_class.__new__, proxy_class)
         else:
             proxy_class_new = None
 
-        @cython.locals(pos=int)
+        @cython.locals(pos = int)
```
tzulberti-jampp, line 4710: Based on PEP8, I think that the original expression was ok

doc/source/CHANGELOG.rst
```diff
@@ -7,6 +7,15 @@ The format is largely inspired by keepachangelog_.
 
 .. _0.1.1:
 
+v0.6.5 - Unreleased
```
tzulberti-jampp: I think this is a new feature so it should be added to version 0.7.0. If we do that, we are going to update libjampp

tzulberti-jampp (commented):
Also bump the version on setup.py and VERSION file

sharedbuffers/mapped_struct.py
llogiudice: Oh, right. It auto-corrected when I was modifying some code. I guess it shouldn't matter (and it's also more consistent with the rest of the code).

doc/source/CHANGELOG.rst
llogiudice: Sure.

tzulberti-jampp (approved):

sharedbuffers/mapped_struct.py
```diff
@@ -167,8 +167,8 @@ def _likebuffer(buf):
     """
     Takes a buffer object as parameter and returns a writable object with buffer protocol.
     """
-    if type(buf) is buffer or type(buf) is bytearray or type(buf) is bytes or (
-            isinstance(buf, bytes) or isinstance(buf, ctypes_Array)):
+    if (type(buf) is buffer or type(buf) is bytearray or type(buf) is bytes or
+            isinstance(buf, (ctypes_Array, bytes))):
```
klaussfreire, line 171: bytes first?

PR #80, opened 2019-06-05
Implement atomic transactions

llogiudice:
This PR implements both atomic compare-and-swap and atomic add on numeric proxy properties. When cython is not being used, these methods are not atomic (And they can't be, unless we use kernel-level synchronization).

Memory order is implied to be sequentially consistent due to the usage of gcc's __sync* builtins.

### DIFF
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -3587,7 +3587,7 @@ class ULongBufferProxyProperty(BaseBufferProxyProperty):
         if cython.compiled:
             _c_buffer_proxy_atomic_add[cython.ulonglong](self, obj, value)
         else:
-            _buffer_proxy_add(self, obj, 'B', value)
+            _buffer_proxy_add(self, obj, 'Q', value)
 
 @cython.cclass
 class LongBufferProxyProperty(BaseBufferProxyProperty):
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -3258,11 +3258,12 @@ if cython.compiled:
         mfence_full()   # release
 
     @cython.cfunc
+    @cython.inline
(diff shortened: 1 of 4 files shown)


sharedbuffers/mapped_struct.py
```diff
@@ -3267,6 +3306,20 @@ def _buffer_proxy_set(self, obj, code, elem):
...
             struct.pack_into(code, obj.buf, obj.offs + self.offs, elem)
 
+    def _buffer_proxy_cas(self, obj, code, exp_val, new_val):
+        # XXX: This is not atomic!
+        if obj is not None and not (obj.none_bitmap & self.mask):
+            tmp = struct.unpack_from(code, obj.buf, obj.offs + self.offs)[0]
+            if tmp == exp_val:
+                struct.pack_into(code, obj.buf, obj.offs + self.offs, new_val)
+                return True
+            return False
```
klaussfreire: This `return False` should be dedented one level I think (all similar ones in fact)

sharedbuffers/mapped_struct.py
```diff
@@ -3286,6 +3339,20 @@ def __set__(self, obj, elem):
...
 
+    @cython.locals(obj = BufferProxyObject, exp_val = cython.uchar, new_val = cython.uchar)
+    def cas(self, obj, exp_val, new_val):
+        if cython.compiled:
+            return _c_buffer_proxy_atomic_cas[cython.uchar](self, obj, exp_val, new_val)
+        else:
+            return _buffer_proxy_cas(self, obj, 'B', exp_val, new_val)
+
+    @cython.locals(obj = BufferProxyObject, value = cython.uchar)
+    def add(self, obj, value):
```
klaussfreire: Add for booleans doesn't make sense, methinks.

klaussfreire (approved):

gefarion (approved):

PR #83, opened 2019-06-21
Smaller RTTI integers

klaussfreire:
Fix mapped_object to actually pick the smallest integer typecode possible
instead of always picking 'q'. Also fix it for 'Q', which wasn't even
working when the number didn't fit in 'q'.


sharedbuffers/mapped_struct.py
```diff
@@ -3005,22 +3005,55 @@ def register_subclass(cls, typ, supertyp):
...
+
+        if typ is None:
+            typ = type(value)
+
+        self.value = value
+
+        if typ is int or typ is long:
+            # Special case for numbers, to pick a smaller typecode when possible
+            try:
+                ivalue = value
```
llogiudice, line 3023: shouldn't this also set the typecode to 'q' ?

klaussfreire: That's done in the `else` part. This particular line is only trying to cast into a signed int64, but success here is not the criteria to actually assign typecode `q`.

llogiudice (approved):

klaussfreire:
Ping

ghost (approved):

PR #84, opened 2019-06-25
Add documentation about binary format and other internals

klaussfreire:


doc/source/internals/ondisk/objects.rst
```diff
@@ -0,0 +1,302 @@
...
+            uint32 ui32;
+            int64 i64;
+            uint64 ui64;
+            float f;
+            double d;
+            bool b;
+        }
+    }
+
+As can be gleaned from the above, the value itself won't be aligned except for the byte-sized types, and
```
llogiudice: This doesn't match the C code. A `union` has the needed size and alignment to hold any of its constituent members.

doc/source/internals/ondisk/sequences.rst
```diff
@@ -0,0 +1,402 @@
...
+
+    union VarLength {
+        // For typecodes l, L, d, t, T
+        long longlength:56;
+
+        // For other typecodes, length < 0xFFFFFF
+        int shortlength:24;
+
+        struct {
+            int padding:24; // 0xFFFFFF
```
llogiudice, line 18: What purpose does this `padding` member serve ?

llogiudice:
I'm not really fond of using `long long`, `unsigned long long` et al to refer to fixed-size C-like integers. Since their sizes aren't standardized, it can lead to confusion when using them in specific platforms or toolchains. I'd much rather use the nomenclature defined in `<stdint.h> `, much like numpy does.

doc/source/internals/ondisk/sequences.rst
klaussfreire: It helps identify long lengths when reading the buffer. Think of this as rice coding.

doc/source/CHANGELOG.rst
```diff
@@ -5,7 +5,14 @@ All notable changes to this project will be documented here.
 
 The format is largely inspired by keepachangelog_.
 
-.. _0.1.1:
+v0.7.1 - Unreleased
```
tzulberti-jampp: I don't think this is a bugfix

doc/source/internals/ondisk/datetime.rst
```diff
@@ -0,0 +1,37 @@
...
+
+    ts = (int(timestamp) << 20) + frac(timestamp) * 1000000
+
+Examples
+--------
+
+.. code-block:: pycon
+
+    >>> buf = bytearray(1 << 20)
+    >>> end = mapped_datetime.pack_into(datetime.datetime(2019, 03, 05, 11, 15, 24, 388001), buf, 0)
```
tzulberti-jampp, line 27: Check the `03` and `05`. They are octal values. Is that required?

doc/source/CHANGELOG.rst
klaussfreire: It isn't. But it's not going to be released on its own either. The final version will be assigned whenever something release-worthy gets pushed. This is more like a pre-emptive version bump.

klaussfreire: In fact, I think this is conflicting with master by now, and by now master is on 0.8.0, illustrating the point.

doc/source/internals/ondisk/datetime.rst
klaussfreire: It's just an exaple. It's probably not necessary, but it was pasted from a console, so I'm sure it works.

klaussfreire:
Ping @tzulberti-jampp @llogiudice and friends

PR #85, opened 2019-07-31
Avoid hashing if key's already hashed

ghost:
I'm new with this so be kind 😄


doc/source/CHANGELOG.rst
```diff
@@ -7,6 +7,15 @@ The format is largely inspired by keepachangelog_.
...
 .. _0.1.1:
 
+v0.8.0 - Unreleased
+===================
+
+Changed
+-------
+
+- ApproxStringIdMultiMapper no longer expects a string as a key in value,
+  allowing already hashed keys (integers)
```
klaussfreire: This reads as if it no longer accepted string keys. Maybe rephrase it to make it clearer that you can still use strings.

tests/mapped_struct.py
```diff
@@ -974,6 +974,8 @@ def gen_values(self, *p, **kw):
         str_ = str
         for k, v in super(ApproxStringIdMultiMapperTest, self).gen_values(*p, **kw):
             yield str_(k), v
+        for k, v in super(ApproxStringIdMultiMapperTest, self).gen_values(*p, **kw):
+            yield k, v
```
klaussfreire: Maybe use a separate test for this. I'm not fond of mixing the two key sets in the same test, it may muddy the test's results.

klaussfreire (changes requested):

klaussfreire (changes requested):
You also forgot to bump the version in setup.py

tests/mapped_struct.py
```diff
@@ -986,6 +986,9 @@ def _testBuild(self, N, tempdir, **gen_kwargs):
                 self.assertIsNotNone(elem)
             if v not in elem:
                 self.assertIn(v, elem)
+            int_elem = rvget(k)
+            if int_elem is None:
+                self.assertIsNotNone(int_elem)
```
klaussfreire: This looks weirdly contradictory.

I'm sure you wanted something else.

Also, it shouldn't be on the base class. You have to do this in `ApproxIntegerIdMultiMapperTest` only.

klaussfreire (changes requested):

tests/mapped_struct.py
ghost: I was just following the previous tests, IDK why it was done that way (check the previous lines, it's the same and the assert above that too)

sharedbuffers/mapped_struct.py
```diff
@@ -8056,7 +8056,10 @@ def build(cls, initializer, *p, **kw):
         encode = cls.encode
         def wrapped_initializer():
             for key, value in initializer:
-                yield xxh(encode(key)).intdigest(), value
+                if isinstance(key, (int, long)):
```
llogiudice, line 8060: You need to update the doc string of the method to reflect the fact that it now accepts integers as keys, too.

ghost: This has docstrings, oh my. Not used to that! haha . will do 👌

tests/mapped_struct.py
```diff
@@ -987,21 +987,33 @@ def _testBuild(self, N, tempdir, **gen_kwargs):
...
-            int_elem = rvget(k)
-            self.assertIsNotNone(int_elem)
-
     # Too much memory
     testBuildHugeInMemShuffled = None
     testBuildHugeOnDiskShuffled = None
 
-class ApproxStringId32MultiMapperTest(ApproxStringIdMultiMapperTest):
+class ApproxStringId32MultiMapperTest(IdMapperTest):
     IdMapperClass = mapped_struct.ApproxStringId32MultiMapper
```
klaussfreire, line 995: What's up with this? This is wrong, you're changing an existing and working test

klaussfreire (changes requested):

tests/mapped_struct.py
ghost: Ups.

klaussfreire (approved):

llogiudice (approved):

PR #88, opened 2019-09-12
Accurate NumericIdMapper hints

klaussfreire:
Implement accurate hints for NumericIdMapper and derivatives,
finally getting improved access patterns for all id mappers.


doc/source/CHANGELOG.rst
```diff
@@ -5,6 +5,15 @@ All notable changes to this project will be documented here.
 
 The format is largely inspired by keepachangelog_.
 
+v0.8.2 - Unreleased
```
tzulberti-jampp, line 8: Why patch the patch version if this isn't a bugfix. Shouldn't we patch de minor?

llogiudice (approved):

doc/source/CHANGELOG.rst
klaussfreire: It's completely transparent and backward compatible as it's only performance. The performance issue could be considered a bug if you squint a bit.

klaussfreire: It also avoids having to patch FaultClient

tzulberti-jampp: I believe this is the real reason...

tzulberti-jampp (approved):

PR #91, opened 2021-01-29
Fix object alignment

klaussfreire:
Fix object attribute alignment by enlarging field bitmaps to match alignment.

Fixes support for ARM architectures that really depend on alignment, but breaks
compatibility with buffers generated by older versions.

Sadly there's only a minimal backward compatibility mechanism in place so it's a binary
compatibility break. Specifically, pickled schemas will contain a version
number that will enable reading old buffers from pickled schemas, but for this
to work all object references must be explicitly typed to allow all schemas
to be pickled properly. If at all possible, it's safer to rebuild all buffers
with 0.9.0 after the update.


valchulen (commented):
Why does making the types signed fix the alignment? Aren't signed and unsigned types the same in bit length?

klaussfreire:
> Why does making the types signed fix the alignment? Aren't signed and unsigned types the same in bit length?

It's not alignment, but arm64 seems to have unsigned chars by default (ie: unqualified char is unsigned), which breaks things since we had assumed signed chars.

benavento (approved):

valchulen (approved):

PR #94, opened 2021-10-20
CORE-4190: Support Python3 & fix pure python tests

chpedersen:
* Most changes are trivial:
  - `"foo"` -> `b"foo"`, `bytes/str/unicode`
  - `iteritems`/`itervalues`/`range`/`xrange` etc.
* In python3 there's no `buffer` so I used `memoryview`. The buffer constructor takes offset & size args, but memoryview doesn't, so I use the `buffer_with_offset` func.
* Some blocks are repeated several times:
```
if six.PY3:
    dtype = obj_dtype.char.encode()
    if cython.compiled:
        cdtype = dtype[0]
    else:
        cdtype = bytes([dtype[0]])
else:
    dtype = obj_dtype.char
    cdtype = cython.cast('const char*', dtype)[0]
```
maybe it's just more efficient to keep them rather than move this to a separate function, IDK
* I don't remeber exactly how, but dict comparison is no the same in py3 as in py2, so I stole the `dict_cmp` from https://stackoverflow.com/questions/3484293/is-there-a-description-of-how-cmp-works-for-dict-objects-in-python-2
* I didn't do big changes in tests, just the "easy" ones. All tests pass :raised_hands: BUT The tests in python3 are slower than in python2, even with cython. In particular, 
`MappedStringMappingBigTest` took about 5 minutes vs 1.5 seconds in python2 :open_mouth: 

Sorry for the not very descriptive 100+ commit messages


klaussfreire:
>     * I didn't do big changes in tests, just the "easy" ones. All tests pass raised_hands BUT The tests in python3 are slower than in python2, even with cython. In particular,
>       `MappedStringMappingBigTest` took about 5 minutes vs 1.5 seconds in python2 open_mouth

That's not a small thing. That's a sign of a serious problem somewhere.

tests/mapped_struct.py
```diff
@@ -855,11 +862,11 @@ def gen_values(self, n, reversed = False, shuffled = False, gen_dupes = False):
             r.shuffle(keys)
         if gen_dupes:
             return itertools.chain(
-                itertools.izip(keys, xrange(0, 2*n, 2)),
-                itertools.islice(itertools.izip(keys, xrange(0, 2*n, 2)), 10, None),
+                zip(keys, xrange(0, 2*n, 2)),
+                itertools.islice(zip(keys, xrange(0, 2*n, 2)), 10, None),
```
klaussfreire, line 866: This materializes the sequences in py2 and will use a lot of RAM, use izip in py2:

``` python
from future.builtins import zip
```

That requires future, but it's a good way to get conisstent behavior.

There's also `map` BTW.

tests/mapped_struct.py
```diff
@@ -1007,9 +1014,9 @@ def _testBuild(self, N, tempdir, **gen_kwargs):
             elem = rv.get(k)
             if isinstance(elem, list):
                 for e in elem:
-                    self.assertTrue(isinstance(e, (int, long)))
+                    self.assertTrue(isinstance(e, (int, long, numpy.uint32)))
```
klaussfreire, line 1017: No, we can't allow numpy types here.

If this is for pure python tests, I'd say rollback. We can require Cython.

tests/mapped_struct.py
```diff
@@ -1094,7 +1101,7 @@ def testBuildValueDedup(self):
         if isinstance(test_values, dict):
             test_values = test_values.items()
```
klaussfreire:
```suggestion
            test_values = iteritems(test_values)
```

tests/mapped_struct.py
```diff
@@ -1292,17 +1299,17 @@ class MappedApproxString32MultiMappingTest(MappedApproxStringMultiMappingTest):
     IdMapperClass = mapped_struct.ApproxStringId32MultiMapper
 
 class MappedStringMappingUnicodeTest(MappedStringMappingTest):
-    TEST_KEYS = [ "%s€ ···YEAH···" % (k,) for k in MappedStringMappingTest.TEST_KEYS ]
+    TEST_KEYS = [b'%s\xe2\x82\xac \xc2\xb7\xc2\xb7\xc2\xb7YEAH\xc2\xb7\xc2\xb7\xc2\xb7' % (k,) for k in MappedStringMappingTest.TEST_KEYS ]
```
klaussfreire: Unreadable.

Use `u"%s€ ···YEAH···".encode("utf8")` maybe?

tests/mapped_struct.py
```diff
@@ -1314,10 +1321,10 @@ class BsearchTest(unittest.TestCase):
 
     for dtype in SUPPORTED_DTYPES:
         def testBsearch(self, dtype=dtype):
-            testarray = range(1,101)
+            testarray = list(range(1,101))
             random.shuffle(testarray)
             a = numpy.array(testarray[:50], dtype)
-            b = numpy.array([0] + testarray[50:] + range(101,103), dtype)
+            b = numpy.array([0] + testarray[50:] + list(range(101,103)), dtype)
```
klaussfreire, line 1327:
``` python
from past.builtins import range, xrange
```

And then you don't have to change these and it works as expected in all cases.

FTR, past and future helpers are only good for tests and pure python code. Cythonized code can't benefit from these because the wrappers aren't recognized in optimizations, so it generates horrible code for their use.

tests/mapped_struct.py
```diff
@@ -1731,7 +1771,7 @@ def testMixed(self):
         self.assertTrue(big >= small)
         self.assertFalse(big > big)
 
-        big = self.pack([1L, 2L, 3L])
+        big = self.pack([1, 2, 3])
```
klaussfreire: They need to be longs. Use `long(1)` maybe as an easy way to make it dual python?

tests/mapped_struct.py
```diff
@@ -2082,10 +2122,24 @@ def testProxiedTupleRepr(self):
...
+        return cmp(type(a).__name__, type(b).__name__)
+    if isinstance(a, tuple):
+        return tuple_cmp(a, b)
+    return cmp(a, b)
 
 class DictPackingCommonTest(object):
 
     def assertUnsortedEquals(self, a, b):
+        if six.PY3:
+            return sorted(a, key=functools.cmp_to_key(key_cmp)) == sorted(b, key=functools.cmp_to_key(key_cmp))
```
klaussfreire, line 2142: I don't follow why you need this. Sorting item tuples should work fine in both python versions.

tests/mapped_struct.py
```diff
@@ -2187,11 +2241,11 @@ def testHashIntegersFixed(self):
             (20913041029, 20913041029),
             (66210231110, 66210231110),
             (752, 752),
-            (-1, 18446744073709551615L),
+            (-1, 18446744073709551615),
```
klaussfreire: Again, use the long(x) idiom

klaussfreire (changes requested):
Partial review only of tests (seemed simpler to start with)

tests/mapped_struct.py
chpedersen: No, it's comparing int and str which doesn't work in py3. For example, from one of the values in MappedDictPackingTest:

in py2
```
sorted(['a', 1, frozenset(), (1, 2)]) == [1, frozenset([]), 'a', (1, 2)]
```
in py3
```
>>> sorted(['a', 1, frozenset(), (1, 2)])
TypeError: '<' not supported between instances of 'int' and 'str'
```

tests/mapped_struct.py
chpedersen: There's the `from six.moves import xxx` for this

sharedbuffers/mapped_struct.py
```diff
@@ -438,7 +438,7 @@ def shared_id(obj):
             return obj | LONG_MASK
         elif isinstance(obj, long):
             # Real longs must make room for the flag bits
-            return (obj & 0xFFFFFFFFFFFFFFFF) | ((obj >> 64) << 68) | LONG_MASK
+            return (obj & long(0xFFFFFFFFFFFFFFFF)) | ((obj >> 64) << 68) | LONG_MASK
```
klaussfreire: This will probably create a python int, not native code. Have you checked the output of `cython --annoate`?

sharedbuffers/mapped_struct.py
```diff
@@ -742,15 +742,15 @@ def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0,
                         # inline unsigned ints
                         dtype = b'I'
                         buf[offs] = ord(dtype)
-                    elif (cython.cast(cython.longlong, -0x8000000000000000) <= iminval
-                            and imaxval <= cython.cast(cython.longlong, 0x7FFFFFFFFFFFFFFF)):
+                    elif (cython.cast(cython.longlong, long(-0x8000000000000000)) <= iminval
+                            and imaxval <= cython.cast(cython.longlong, long(0x7FFFFFFFFFFFFFFF))):
```
klaussfreire: Idem. Why this change? It should not have been affected by py3, and this way won't be optimal code IIUC.

sharedbuffers/mapped_struct.py
```diff
@@ -125,13 +125,16 @@
     long = int
     basestring = (bytes, str)
 
+def make_memoryview(a):
+    return memoryview(a).cast("B")
```
klaussfreire: Why don't you make it a cfunc, inline, and put hte `if six.PY2` check inside?

Make sure to cache the `six.PY2` bit in a native global to avoid slow python access inside it.

sharedbuffers/mapped_struct.py
```diff
@@ -141,7 +144,7 @@
     assert Py_GE == 5
 else:
     if six.PY3:
-        buffer = memoryview
+        buffer = make_memoryview
```
klaussfreire: In py2, memoryview has no `cast`, so this doesn't work as-is.

sharedbuffers/mapped_struct.py
```diff
@@ -967,7 +967,10 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, klass = list)
             if objlen == 0xFFFFFF:
                 objlen = _struct_l_Q.unpack_from(buf, offs)
                 offs += 8
-            rv = array(dtype, buf[offs:offs+itemsize*objlen])
+            w = buf[offs:offs+itemsize*objlen]
+            if six.PY3:
+                w = bytes(w)
```
klaussfreire: This will copy

klaussfreire (changes requested):

tests/mapped_struct.py
```diff
@@ -12,20 +12,27 @@
 from datetime import datetime, date
 from decimal import Decimal
 import numpy as np
+import six
+
+if six.PY3:
+    long = int
+    buffer = memoryview
```
olifer97, line 19: this is because we will be using this library in python2 and python3 right?

tests/mapped_struct.py
olifer97: why not numpy types?

tests/mapped_struct.py
```diff
@@ -1699,6 +1706,39 @@ def testSubsetSuperset(self):
...
+        self.assertFalse(b <= a)
+        self.assertNotEquals(a, b)
+
+        a = self.pack([1., 2., 3., 10.])
+        b = self.pack([1., 2., 3., 4.])
+        self.assertFalse(a < b)
+        self.assertFalse(a <= b)
+        self.assertFalse(b < a)
+        self.assertFalse(b <= a)
+        self.assertNotEquals(a, b)
```
olifer97, line 1739: why do we have to add all this asserts after migrating to python3? the behavior shouldn't be the same?

olifer97 (commented):
The idea is to have this library support both python2 and python3 because we dont have all the components that used this library migrated right?
I dont think i have enough knowledge to approved this, the problem with the duration of tests seems like a severe problem if we use this library in places perfomance-sensitive.

tests/mapped_struct.py
klaussfreire: Oh... it's a mixed-type test, didn't notice.

What about `key=repr`?

sharedbuffers/mapped_struct.py
klaussfreire: BTW... the `long(x)` idiom is only for tests because of this (inefficient). It's ok in tests, we don't care, but actual library code should be efficient.

tests/mapped_struct.py
klaussfreire: The original input were native types. If the library produces numpy types on output, something is broken.

tests/mapped_struct.py
klaussfreire: Hadn't seen it

sharedbuffers/mapped_struct.py
```diff
@@ -462,30 +492,43 @@ def is_wrapped_key(obj):
...
-                if cdtype in ('l', 'I', 'i', 'H', 'h', 'B', 'b'):
+                if six.PY3:
+                    dtype = obj_dtype.char.encode()
+                    if cython.compiled:
+                        cdtype = dtype[0]
+                    else:
+                        cdtype = bytes([dtype[0]])
+                else:
+                    dtype = obj_dtype.char
+                    cdtype = cython.cast('const char*', dtype)[0]
```
klaussfreire: You can't have this check (six.py3) in performance-critical code, as it compiles to a python attribute access on the global module. A relatively slow operation.

Store the result of this check into a cython.bint global variable, and use that variable instead if you need to.

Otherwise, try to write version-agnostic code.

In this case, if you declare dtype as unicode, `dtype[0] in ('l', 'I', 'i', ...)` generates pretty good code if you take a look. You would need 2 verisons of the code, py2 and py3, but the py3 version would actually be quite simpler (no char pointer stuff).

sharedbuffers/mapped_struct.py
```diff
@@ -575,22 +618,28 @@ def unpack_from(cls, buf, offs, idmap = None):
         else:
             pbuf = buf
         try:
-            fs_type = pbuf[offs]
-            if fs_type == 'm' or fs_type == 'M':
+            if six.PY3 and not cython.compiled:
```
klaussfreire: cython.compiled is a static check, so always do it first, as it's the fastest

sharedbuffers/mapped_struct.py
```diff
@@ -81,13 +82,19 @@
 import ctypes
 from datetime import timedelta, datetime, date
 from decimal import Decimal
+from six import itervalues, reraise, iteritems, iterkeys
+from six.moves import cPickle, range
```
klaussfreire: importing range may make cython's loop optimizations go bad.

Check cython --annotate

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
chpedersen: Ok, I see that calling `long` is expensive here. But why did we have the "L" suffix in the first place?  It seems to me it doesn't change anything

```python
cython.cast(cython.longlong, -0x8000000000000000L)
```

turns into
```c
(PY_LONG_LONG)-9223372036854775808L)
```

and 
```python
cython.cast(cython.longlong, -0x8000000000000000)
```
into
```c
(PY_LONG_LONG)-9223372036854775808LL)
```

klaussfreire: Did you try python 2?

klaussfreire: TBH, I think the correct version is the second version, because it produces a LL literal, which is the right kind. But I suspect that's not the case in py2 because we've had issues with that in the past, where it would use a regular int literal that gets truncated.

Truncation produces a warning, so you can check for warnings while building.

klaussfreire: py2:
```
+4: x = cython.cast(cython.longlong, -0x8000000000000000)

  __pyx_v_4test_x = ((PY_LONG_LONG)-9223372036854775808LL);
```

py3
```
+4: x = cython.cast(cython.longlong, -0x8000000000000000)

  __pyx_v_4test_x = ((PY_LONG_LONG)-9223372036854775808LL);
```

klaussfreire: Looks like removing the L suffix is the right way to go about it. Must have been fixed in sme cython patch because it was needed before.

`\_(0_0)_/`

sharedbuffers/mapped_struct.py
chpedersen: `make_memoryview` is only used in py3

sharedbuffers/mapped_struct.py
chpedersen: oops, yes. i'll look into it

chpedersen: We could just use `range` without importing, although in py2 without cython it would be slower than xrange of course (py2 + cython is ok with `range`). I don't see another way around this.

sharedbuffers/mapped_struct.py
klaussfreire: You can't do that with make_memoryview as a cfunc. You have to make make_memoryview work in both versions, and have all the python2 conditionals inside it.

That's because you can't assign a cfunc to a python variable. I'm not sure how it's working for you now in fact. It should break hard.

sharedbuffers/mapped_struct.py
```diff
@@ -441,7 +446,7 @@ def shared_id(obj):
             return obj | LONG_MASK
         elif isinstance(obj, long):
             # Real longs must make room for the flag bits
-            return (obj & long(0xFFFFFFFFFFFFFFFF)) | ((obj >> 64) << 68) | LONG_MASK
+            return (obj & cython.cast(cython.longlong, 0xFFFFFFFFFFFFFFFF)) | ((obj >> 64) << 68) | LONG_MASK
```
klaussfreire: longlong can't handle that literal. It needs an ulonglong, but since you're doing `obj &` you really need a python long.

sharedbuffers/mapped_struct.py
```diff
@@ -745,15 +750,15 @@ def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0,
...
+                    elif (cython.cast(cython.longlong, -0x8000000000000000) <= iminval
+                            and imaxval <= cython.cast(cython.longlong, 0x7FFFFFFFFFFFFFFF)):
                         # inline signed int64 list
                         buf[offs] = ord('q')
                         dtype = b'l'
                     else:
                         raise OverflowError
                 except OverflowError:
-                    if 0 <= minval and maxval <= long(0xFFFFFFFFFFFFFFFF):
+                    if 0 <= minval and maxval <= cython.cast(cython.longlong, 0xFFFFFFFFFFFFFFFF):
```
klaussfreire: longlong can't handle that literal. You need an ulonglong, and IIRC using an ulonglong generates ugly code, so a python long is a better choice. I think using simply `0xFFFFFFFFFFFFFFFF` should work in both versions, but do verify with the annotated code.

sharedbuffers/mapped_struct.py
```diff
@@ -103,6 +103,9 @@
 
 import cython
 
+python2 = cython.declare(cython.bint, six.PY2)
+python3 = cython.declare(cython.bint, six.PY3)
```
klaussfreire, line 101: I know it sounds unnecessary, but color me paranoid:

``` python
assert python2 or python3
```

sharedbuffers/mapped_struct.py
```diff
@@ -1377,7 +1382,7 @@ def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0):
         See `mapped_object.pack_into` for argument details.
         """
         cur_offs = offs
-        if six.PY3:
+        if python3:
             objlen = obj.nbytes
```
klaussfreire: I'm not sure about using .nbytes just like that.

Why don't you take a memoryview explicitly to do the cast to bytes?

We want to be polymorphic here and accept byte strings, bytearrays, and iterables of int. Using nbytes straightaway breaks for byte strings and bytearrays both.

In fact, maybe add a test to make sure this works with byte stirngs and byte arrays.

sharedbuffers/mapped_struct.py
```diff
@@ -950,9 +955,9 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, klass = list)
             if objlen == 0xFFFFFF:
                 objlen = _struct_l_Q.unpack_from(buf, offs)
                 offs += 8
-            w = buf[offs:offs+itemsize*objlen]
-            if six.PY3:
-                w = bytes(w)
+            w = buf[offs:offs + itemsize * objlen]
+            if python3:
+                w = memoryview(w).cast(dtype)
```
klaussfreire: Isn't it a memoryview already by now?

It comes from likerobuffer.

sharedbuffers/mapped_struct.py
```diff
@@ -965,7 +970,7 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, klass = list)
             objlen >>= 8
             offs += 8
             q = buf[offs:offs + itemsize * objlen]
-            if six.PY3:
+            if python3:
                 q = bytes(q)
```
klaussfreire: All those bytes(q) will also copy. Those also need fixing.

sharedbuffers/mapped_struct.py
```diff
@@ -143,16 +148,16 @@ def make_memoryview(a):
...
-    if six.PY3:
+    if python3:
         buffer = make_memoryview
 
 def buffer_with_offset(data, offset, size=None):
     if size is not None:
-        if six.PY3:
+        if python3:
             return buffer(data)[offset:offset+size]
         return buffer(data, offset, size)
```
klaussfreire: I feel more fuzzy with an else in there

sharedbuffers/mapped_struct.py
```diff
@@ -492,7 +497,7 @@ def is_wrapped_key(obj):
 @cython.inline
 def get_wrapped_key(obj):
     # the key for the unwrapped value
-    if six.PY2:
+    if python2:
         if isinstance(obj, tuple):
             return obj[1]
         elif isinstance(obj, int):
```
klaussfreire: Again, I prefer else for exclusive paths:

``` python
if python2:
   # py2 verison
else:
   # py3 verison
```

Instead of

``` python
if python3:
    # py3 version
# py2 version
```

It's just so much more readable in the first form.

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
```diff
@@ -514,27 +514,29 @@ def get_wrapped_key(obj):
 class mapped_frozenset(frozenset):
 
     @classmethod
-    @cython.locals(cbuf = 'unsigned char[:]', i=int, ix=int, offs=cython.longlong, cdtype=cython.uchar)
+    @cython.locals(cbuf = 'unsigned char[:]', i=int, ix=int, offs=cython.longlong, py2_dtype=cython.uchar, py3_dtype="unicode")
```
klaussfreire: Instead of py2/3_dtype, cdtype and udtype?

sharedbuffers/mapped_struct.py
```diff
@@ -514,27 +514,29 @@ def get_wrapped_key(obj):
...
-                    dtype = obj_dtype.char
-                    cdtype = cython.cast('const char*', dtype)[0]
-                if cdtype in (b'l', b'I', b'i', b'H', b'h', b'B', b'b'):
-                    all_int = all_intlong = 1
-                elif cdtype == b'L':
-                    all_intlong = 1
-                elif cdtype in (b'd', b'f'):
-                    all_float = 1
+                    temp = obj_dtype.char
+                    py2_dtype = cython.cast('const char*', temp)[0]
```
klaussfreire: Add a comment stating why you need temp (to keep a reference to the string while you access the char*)

Otherwise I can totally see someone thinking it's not necessary and causing issues later on

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
klaussfreire: You can trust cython to properly implement xrange, I think.

Keep unapologetically using xrange, cython will generate proper code for it even if it's not there in py3.

Cython is not optional after all, lets remember that. The fact that tests are failing in pure python is a sign of that.

If you want to cater to pure python, you can do the globals trick:

``` python
if not cython.compiled:
     from six.moves import range as _range
     globals().update(dict(range=_range))
```

That has the same effect as an import in pure python, but is invisible to Cython.

sharedbuffers/mapped_struct.py
```diff
@@ -504,38 +506,38 @@ def get_wrapped_key(obj):
...
-
-    if isinstance(obj, tuple):
-        return obj[1]
-    elif isinstance(obj, int):
-        return obj & ~WRAP_MASK
+    else:
+        if isinstance(obj, tuple):
+            return obj[1]
+        elif isinstance(obj, int):
+            return obj & ~WRAP_MASK
```
klaussfreire: You can simplify this.

The python2 and 3 sides are identical except for the `long` case.

``` python
if isinstance(obj, tuple):
    return obj[1]
elif python2:
    # py2 cases
else:
    # py3 cases
```

That way you avoid repeating the common case.

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
klaussfreire: ping

klaussfreire: You may be able to avoid the `buffer` global and replace all uses of buffer with make_memoryview if you do that BTW. It will work better.

sharedbuffers/mapped_struct.py
```diff
@@ -575,22 +627,28 @@ def unpack_from(cls, buf, offs, idmap = None):
         else:
             pbuf = buf
         try:
-            fs_type = pbuf[offs]
-            if fs_type == 'm' or fs_type == 'M':
+            if not cython.compiled and python3:
+                fs_type = bytes([pbuf[offs]])
+            else:
+                fs_type = pbuf[offs]
```
klaussfreire: This is unnecessary. `pbuf` is an `uchar*` which means `pbuf[offs]` will be an `uchar` regardless of python version.

If this is catering to pure python runs, remember we don't need pure python support. We should drop it and avoid the complications it brings, which means you should revert these to the cython-native implementation.

We should probably switch all `if cython.compiled` into `assert cython.compiled` (in another PR) because we should really drop pure python support. It's way too hard to maintain and completely unnecessary.

Pure python support was meant to allow running natively in PyPy, but PyPy's cpyext improved to the point where it can run cythonized code quite robustly, so we can depend on cython and simplify all those pesky `if cython.compiled` cases.

So... for now, just forget about pure python tests. Worry about making it run in cython.

sharedbuffers/mapped_struct.py
```diff
@@ -716,30 +788,30 @@ def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0,
             if isinstance(obj, npndarray):
                 a = obj
             else:
-                a = array(dtype, obj)
+                a = array(dtype.decode(), obj)
             abuf = buffer(a)
             buf[offs:offs+len(abuf)] = abuf
             offs += len(abuf)
-            offs = (offs + 7) / 8 * 8
+            offs = (offs + 7) // 8 * 8
```
klaussfreire, line 835: The best way to avoid headaches with division semantics in dual code, is to add `from __future__ import division` to the imports on the top, and make sure we always consistently use py3 semantics.

This means using `//` here as you did, but without the future import this will break py2.

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
```diff
@@ -501,17 +503,15 @@ def is_wrapped_key(obj):
 @cython.inline
 def get_wrapped_key(obj):
     # the key for the unwrapped value
+    if isinstance(obj, tuple):
+        return obj[1]
     if python2:
```
```suggestion
    elif python2:
```

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
```diff
@@ -776,7 +788,12 @@ def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0,
                         # longs are tricky, give up
                         all_int = all_intlong = 0
         if all_int or all_intlong:
-            if dtype == b'l' or dtype == b'L':
+            l_or_L = False
+            if check_udtype:
+                l_or_L = udtype[0] == u'l' or udtype[0] == u'L'
+            else:
+                l_or_L = dtype == b'l' or dtype == b'L'
```
klaussfreire, line 816: There are cases in py3 where you can end up here. Wouldn't that fail the comparison because of unicode-bytes mismatch?

Maybe always set udtype and instead of "check_udtype" just use "if python3"

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
```diff
@@ -671,18 +671,18 @@ class mapped_tuple(tuple):
...
         offs = cython.Py_ssize_t, implicit_offs = cython.Py_ssize_t, val_offs = cython.Py_ssize_t,
-        i = cython.Py_ssize_t, iminval = cython.longlong, imaxval = cython.longlong, cdtype=cython.uchar, udtype='unicode', check_udtype=cython.bint)
+        i = cython.Py_ssize_t, iminval = cython.longlong, imaxval = cython.longlong, cdtype=cython.uchar, udtype='unicode', l_or_L=cython.bint)
     def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0,
             array = array.array):
         baseoffs = offs
         objlen = len(obj)
-        check_udtype = False
+        udtype = u''
+        dtype = b''
```
klaussfreire: These are not valid dtypes. I'd rather use None.

sharedbuffers/mapped_struct.py
```diff
@@ -807,7 +817,10 @@ def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0,
             if isinstance(obj, npndarray):
                 a = obj
             else:
-                a = array(dtype.decode(), obj)
+                if python3:
+                    a = array(udtype, obj)
+                else:
+                    a = array(dtype, obj)
```
```suggestion
                a = array(udtype if python3 else dtype, obj)
```

sharedbuffers/mapped_struct.py
klaussfreire: ping

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
chpedersen: > I'm not sure about using .nbytes just like that.

ok, I changed this

> We want to be polymorphic here and accept byte strings, bytearrays, and iterables of int. Using nbytes straightaway breaks for byte strings and bytearrays both.

`proxied_buffer.unpack_from` returns a `buffer`. So the `self.assertEquals(obj, unpacked_obj)` fails if `obj` is a string (bytes) in py2. (in py3 it works, `memoryview(obj) == obj` is true)

klaussfreire: Maybe the test can cast the expected object to buffer in the assert then:

``` python
self.assertEquals(buffer(obj), unpacked_obj)
```

sharedbuffers/mapped_struct.py
klaussfreire: Ping. Add the future import.

chpedersen: It's already there :P

klaussfreire: No... where?

klaussfreire: NVM... I was clicking the file name and github was sending me to an old version of the file. I see it now.

klaussfreire:
I'm getting a failure in python 3.9:

```
======================================================================
FAIL: testProxiedTupleHash (tests.mapped_struct.ProxiedTuplePackingTest)
----------------------------------------------------------------------
  File "/home/claudiofreire/src/sharedbuffers/tests/mapped_struct.py", line 2075, in testProxiedTupleHash
    self.assertTrue(hash(p) == hash(t))
AssertionError: False is not true

----------------------------------------------------------------------
Ran 637 tests in 1901.311s
```

I suspect it has something to do with python using randomized hash seeds.

The tests took also significantly longer than in py2, with py2 doing them all in 1498s and py3 1901s. There's probably something slow in the new code, maybe memory being copied that shouldn't or something like that.

sharedbuffers/mapped_struct.py
```diff
@@ -1111,6 +1236,10 @@ def seek(self, pos):
 _DICT_HEADER_PACKER = cython.declare(object, struct.Struct('=Q'))
 _DICT_HEADER_SIZE = cython.declare(cython.Py_ssize_t, _DICT_HEADER_PACKER.size)
 
+if python3:
+    def cmp(a, b):
+        return (a > b) - (a < b)
+
 @cython.cclass
 class proxied_dict(object):
```
klaussfreire, line 1260: You need to change proxied_dict to have python 3's dict interface: keys, values, items are all iterators now. In fact keys yields a set-like interface, ideally we'd return a proxied_set kind of thing. It will take some doing. Maybe for a future PR, but for now, yielding iterators is good enough.

sharedbuffers/mapped_struct.py
```diff
@@ -114,17 +128,44 @@
...
+else:
+    if python3:
+        buffer = make_memoryview
+
+def buffer_with_offset(data, offset, size=None):
+    if size is not None:
+        if python3:
+            return buffer(data)[offset:offset+size]
+        else:
+            return buffer(data, offset, size)
```
klaussfreire: Should call make_memoryview directly (in both python versions), to make use of it being cdef. Have the py3 conditionals inside make_memoryview. Been saying it for a while.

sharedbuffers/mapped_struct.py
```diff
@@ -6326,7 +6595,14 @@ def build(cls, initializer, destfile = None, tempdir = None,
...
 
             indexpos = curpos
-            write(buffer(index))
+            if python3:
+                if len(index):
+                    w = buffer(index)
+                else:
+                    w = b''
+            else:
+                w = buffer(index)
```
klaussfreire: This does not seem necessary. I'm perfectly able to do a memoryview of an empty array in py3. What exactly prompted this?

sharedbuffers/mapped_struct.py
```diff
@@ -464,29 +505,42 @@ def get_wrapped_key(obj):
...
-                elif cdtype in ('d', 'f'):
-                    all_float = 1
+                if python3:
+                    udtype = obj_dtype.char
+                    if udtype[0] in (u'l', u'I', u'i', u'H', u'h', u'B', u'b'):
+                        all_int = all_intlong = 1
+                    elif udtype[0] == u'L':
+                        all_intlong = 1
+                    elif udtype[0] in (u'd', u'f'):
+                        all_float = 1
```
```suggestion
                    udtype = obj_dtype.char
                    udtype0 = udtype[0]
                    if udtype0 in (u'l', u'I', u'i', u'H', u'h', u'B', u'b'):
                        all_int = all_intlong = 1
                    elif udtype0 == u'L':
                        all_intlong = 1
                    elif udtype0 in (u'd', u'f'):
                        all_float = 1
```

Seems to generate substantially better code

sharedbuffers/mapped_struct.py
klaussfreire: ping. This is breaking cython optimizations. Really needs fixing.

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
chpedersen: It breaks when `index = array([], shape=(0, 2), dtype=uint64)`
```
File "sharedbuffers/mapped_struct.py", line 139, in sharedbuffers.mapped_struct.make_memoryview
    return memoryview(a).cast("B")
TypeError: memoryview: cannot cast view with zeros in shape or strides
```

chpedersen:
> I'm getting a failure in python 3.9:
> 
> ```
> ======================================================================
> FAIL: testProxiedTupleHash (tests.mapped_struct.ProxiedTuplePackingTest)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/claudiofreire/src/sharedbuffers/tests/mapped_struct.py", line 2075, in testProxiedTupleHash
>     self.assertTrue(hash(p) == hash(t))
> AssertionError: False is not true
> 
> ----------------------------------------------------------------------
> Ran 637 tests in 1901.311s
> ```
> 
> I suspect it has something to do with python using randomized hash seeds.

The hash function for tuples was changed: https://github.com/python/cpython/pull/9471/files so I guess we have to translate the new version to python and use that one for py 3.8+. BTW why are we doing this (in `proxied_tuple.__hash__`), instead of just `self._hash = hash(tuple(iter(self)))` like below?

chpedersen:
Since we are now requiring cython, I removed the `not cython.compiled` cases that I added _in this PR_

klaussfreire:
> > I'm getting a failure in python 3.9:
> > ```
> > ======================================================================
> > FAIL: testProxiedTupleHash (tests.mapped_struct.ProxiedTuplePackingTest)
> > ----------------------------------------------------------------------
> > Traceback (most recent call last):
> >   File "/home/claudiofreire/src/sharedbuffers/tests/mapped_struct.py", line 2075, in testProxiedTupleHash
> >     self.assertTrue(hash(p) == hash(t))
> > AssertionError: False is not true
> > 
> > ----------------------------------------------------------------------
> > Ran 637 tests in 1901.311s
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I suspect it has something to do with python using randomized hash seeds.
> 
> The hash function for tuples was changed: https://github.com/python/cpython/pull/9471/files so I guess we have to translate the new version to python and use that one for py 3.8+. BTW why are we doing this (in `proxied_tuple.__hash__`), instead of just `self._hash = hash(tuple(iter(self)))` like below?

Because it's faster. Creating a tuple only to hash it has move overhead.

klaussfreire:
> The hash function for tuples was changed: https://github.com/python/cpython/pull/9471/files so I guess we have to translate the new version to python and use that one for py 3.8+. BTW why are we doing this (in `proxied_tuple.__hash__`), instead of just `self._hash = hash(tuple(iter(self)))` like below?

Beware, there are 2 contexts in which we may require a hash.

To implement `__hash__`, in which case we can do whatever (keep the old or use the new), each option has its pros and cons.

And to use as hashing for the object mapper (stablehash), in which case we **cannot** update it, as it would break binary compatibility.

In general, there's no need to update the algorithm. We can keep the old. In practice, the change in py3 was made for a reason, and we might want to follow suit, but not at the cost of breaking binary compatibility. So I would not change stablehash.

sharedbuffers/mapped_struct.py
```diff
@@ -135,35 +135,35 @@
...
-    if python3:
-        buffer = make_memoryview
 
+@cython.cfunc
+@cython.inline
 def buffer_with_offset(data, offset, size=None):
     if size is not None:
         if python3:
-            return buffer(data)[offset:offset+size]
+            return memoryview(data).cast("B")[offset:offset+size]
```
klaussfreire: You should probably call make_memoryiew to centralize the logic there

sharedbuffers/mapped_struct.py
```diff
@@ -135,35 +135,35 @@
 @cython.cfunc
 @cython.inline
 def make_memoryview(a):
-    return memoryview(a).cast("B")
+    if python3:
+        return memoryview(a).cast("B")
```
klaussfreire: Can we avoid creating the memoryview?

``` python
if type(a) is not memoryview:
    a = memoryview(a)
```

There's probably a way to also avoid the cast... it should be a fast way, `a.format == "B"` works but it may be slow if it's python operations instead of native. I could not find a way to make it native.

I did a test, and it seems even without cython help it's quite faster to avoid the unnecessary operations:

``` python
import cython
  
_memoryview = cython.declare(object)
_memoryview = memoryview

def fbuffer(a):
    rv = a if type(a) is _memoryview else _memoryview(a)
    if rv.format != 'B':
        rv = rv.cast('B')
    return rv

def mbuffer(a):
    rv = a if type(a) is _memoryview else _memoryview(a)
    return rv.cast('B')

def sbuffer(a):
    return _memoryview(a).cast('B')
```

With a few timeits:

```
>>> import timeit, cython, pyximport
>>> pyximport.install(pyimport=True)
(<pyximport.pyximport.PyImporter object at 0x7ff42a518cc0>, <pyximport.pyximport.PyxImporter object at 0x7ff42a51ec18>)
>>> import ctest
>>> timeit.timeit(lambda:ctest.fbuffer(b"hola"), number=10000000) / 10.0
0.3661114629008807
>>> timeit.timeit(lambda:ctest.mbuffer(b"hola"), number=10000000) / 10.0
0.5427464027947281
>>> timeit.timeit(lambda:ctest.sbuffer(b"hola"), number=10000000) / 10.0
0.5551418196002487
>>> timeit.timeit(lambda:ctest.fbuffer(ctest.fbuffer(b"hola")), number=10000000) / 10.0
0.5739469211956021
>>> timeit.timeit(lambda:ctest.mbuffer(ctest.mbuffer(b"hola")), number=10000000) / 10.0
0.8258275890024379
>>> timeit.timeit(lambda:ctest.sbuffer(ctest.sbuffer(b"hola")), number=10000000) / 10.0
0.9486311056010891
```

It's clear that fbuffer wins by a healthy margin. It's a small difference (a fraction of a microsecond), but it's used a lot this make_memoryview thing, so I think it's worth paying attention to performance there.

sharedbuffers/mapped_struct.py
```diff
@@ -1309,19 +1301,28 @@ def iteritems(self):
...
+        if python3:
+            return iter(self.objmapper)
+        else:
+            return self.objmapper.keys()
 
     def itervalues(self):
         return iter(self.vlist)
 
     def values(self):
-        return self.vlist
```
klaussfreire: Why not leave it as is? The vlist is iterable, so it's a valid return value.

sharedbuffers/mapped_struct.py
```diff
@@ -1309,19 +1301,28 @@ def iteritems(self):
...
+        else:
+            return list(self.iteritems())
 
     def iterkeys(self):
         return iter(self.objmapper)
 
     def keys(self):
-        return self.objmapper.keys()
+        if python3:
+            return iter(self.objmapper)
```
klaussfreire: Why not just 

```suggestion
            return self.objmapper
```

It's iterable, so it's a valid return value (py3 returns an iterable set-like object not a pure iterator)

We're missing the set-iike part but maybe we can fix it in objmapper in some future PR.

sharedbuffers/mapped_struct.py
```diff
@@ -6595,13 +6546,10 @@ def build(cls, initializer, destfile = None, tempdir = None,
...
-                if len(index):
-                    w = buffer(index)
-                else:
-                    w = b''
+            if python3 and len(index) == 0:
+                w = b''
             else:
-                w = buffer(index)
+                w = make_memoryview(index)
             write(w)
```
klaussfreire: What about

``` python
nitems = len(index)
if nitems != 0:
    write(make_memoryview(index))
```

sharedbuffers/mapped_struct.py
```diff
@@ -7232,11 +7180,11 @@ def build(cls, initializer, destfile = None, tempdir = None, return_mapper = Tru
...
         if python3:
             if len(index):
-                d = buffer(index)
+                d = make_memoryview(index)
             else:
                 d = b''
         else:
-            d = buffer(index)
+            d = make_memoryview(index)
         write(d)
```
klaussfreire: Similarly here, I think it can be simplified.

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
```diff
@@ -4394,7 +4628,7 @@ def get_packer(self, obj):
             ]))
```
klaussfreire, line 4662: I'm getting these errors in tests:

```
======================================================================
ERROR: testUnpack (tests.pool.SmallIntContainerPackingTest)
----------------------------------------------------------------------
  File "/home/claudiofreire/src/sharedbuffers/tests/pool.py", line 114, in testUnpack
    pos = p.pack(self.schema, x)[0]
  File "/home/claudiofreire/src/sharedbuffers/sharedbuffers/pool.py", line 225, in pack
    pos = self._pack_into(schema, obj, section)
  File "/home/claudiofreire/src/sharedbuffers/sharedbuffers/pool.py", line 171, in _pack_into
    buf = schema.pack(obj, idmap, implicit_offs=implicit_offs)
  File "sharedbuffers/mapped_struct.py", line 4809, in sharedbuffers.mapped_struct.Schema.pack
    def pack(self, obj, idmap = None, packer = None, padding = None, implicit_offs = 0):
  File "sharedbuffers/mapped_struct.py", line 4819, in sharedbuffers.mapped_struct.Schema.pack
    endp = self.pack_into(obj, buf, 0, idmap, packer, padding, implicit_offs)
  File "sharedbuffers/mapped_struct.py", line 4774, in sharedbuffers.mapped_struct.Schema.pack_into
    packer, padding = self.get_packer(obj)
  File "sharedbuffers/mapped_struct.py", line 4622, in sharedbuffers.mapped_struct.Schema.get_packer
    packer = struct.Struct("".join([
TypeError: sequence item 0: expected str instance, bytes found
```

Seems Struct.format is bytes, and the join there is using strings. Changing it to `b"".join` should fix it. There's quite a few places like these.

Edit: seems to affect python 3.6, not 3.9. Don't know about the in-between versions.

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
klaussfreire: Yep, seems 3.6 has bytes in Struct.format, but not the other versions. Just doing b"".join doesn't fix it for all, it breaks in 3.7+.

Checking if there's a way that's worth the trouble. Otherwise we may have to drop support for 3.6-

klaussfreire: Ok, it looks ugly, but this seems to work, and it's a small effort to get 3.6 support as it's only in one place, so might as well do it:

``` diff
diff --git a/sharedbuffers/mapped_struct.py b/sharedbuffers/mapped_struct.py
index 41f19ce..51f4df5 100644
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -4618,9 +4618,12 @@ class Schema(object):
             packer_key = present_bitmap
 
         rv = self.packer_cache.get(packer_key)
+        bitmap_packer_format = self.bitmap_packer.format
+        if isinstance(bitmap_packer_format, bytes):
+            bitmap_packer_format = bitmap_packer_format.decode('ascii')
         if rv is None:
             packer = struct.Struct("".join([
-                self.bitmap_packer.format,
+                bitmap_packer_format,
             ] + [
                 self.slot_struct_types[slot]
                 for i,slot in enumerate(self.slot_keys)
```

chpedersen: Yep, I was just doing something like this

chpedersen:
@klaussfreire

tests/mapped_struct.py
```diff
@@ -110,8 +117,8 @@ class PrimitiveStruct(TestStruct):
     __slot_types__ = {
         'a' : int,
         'b' : float,
-        's' : str,
-        'u' : unicode,
+        's' : bytes,
+        'u' : six.text_type,
```
klaussfreire, line 121: I prefer importing text_type as unicode, ends up bytes/unicode which is clear and unambiguous.

tests/mapped_struct.py
```diff
@@ -2202,29 +2256,30 @@ def testHashFloatsFixed(self):
...
         )
         self.assertHashesOK(values_hashes)
 
     def testHashSequenceFixed(self):
         values_hashes = (
-            ((1, "abc", -2.3), 2059039662577021167L),
-            (frozenset([1.2, 4.5, 0.12]), 4015877951310865576L),
+            ((1, "abc", -2.3), long(2059039662577021167)),
+            # FIXME this doesnt work, the frozensets have differente order in py3 vs py2
+            # (frozenset([1.2, 4.5, 0.12]), 4015877951310865576),
```
klaussfreire, line 2282: This is a problem. This test is testing stable_hash, and as the name implies, it should be stable and not order dependent.

sharedbuffers/mapped_struct.py
```diff
@@ -81,13 +84,19 @@
 import ctypes
 from datetime import timedelta, datetime, date
 from decimal import Decimal
+from six import itervalues, reraise, iteritems, iterkeys
```
klaussfreire: As I mentioned somewhere else, you shouldn't use them in cythonized code. Cython does not recognize them and thus doesn't optimize them. We want optimized iteration in most cases.

sharedbuffers/mapped_struct.py
```diff
@@ -114,10 +128,23 @@
 
 ctypes_Array = cython.declare(object, ctypes.Array)
 
+if python3:
+    long = int
+    basestring = (bytes, str)
```
``` python
from past.builtins import basestring
```

But, again, these break cython's ability to perform optimizations.

I think it's best to trust cython. Cython understands unicode and bytes in both versions properly without any magic compatibility declarations.

I could only get accelerated isinstance checks by using `isinstance(x, (bytes, unicode))` literally, without any helper constant. You may have to do it that way when necessary.

Cython also understands basestring, but in py3 it becomes just str (unicode), does not include bytes. In some contexts, it may be correct, so pay attention to context.

klaussfreire (changes requested):
Check those six/past uses in mapped_struct, they inhibt optimizations.

In the meanwhile I'll give it a thorough inspection with cython.annotate to look for other issues.

sharedbuffers/mapped_struct.py
```diff
@@ -81,13 +84,19 @@
...
 try:
     from cdecimal import Decimal as cDecimal
 except:
     cDecimal = Decimal
 
-from chorde.clients.inproc import Cache
+try:
+    from chorde.clients.inproc import Cache
+except ImportError:
+    from clru.lrucache import LRUCache as Cache
```
klaussfreire: Should not attempt to import chorde at all. Otherwise it still breaks with SyntaxError if installed from the py2 branch.

sharedbuffers/mapped_struct.py
```diff
@@ -2395,7 +2609,15 @@ def __repr__(self):
...
-is_cpython = cython.declare(cython.bint, sys.subversion[0] == 'CPython')
+if python3:
+    _cpython = sys.implementation.name == 'cpython'
+else:
+    _cpython = sys.subversion[0] == 'CPython'
+is_cpython = cython.declare(cython.bint, _cpython)
+
+XXPRIME_1 = cython.declare(cython.long, 11400714785074694791)
+XXPRIME_2 = cython.declare(cython.long, 14029467366897019727)
+XXPRIME_5 = cython.declare(cython.long, 2870177450012600261)
```
klaussfreire: I think those are longlong

sharedbuffers/mapped_struct.py
```diff
@@ -2412,24 +2634,49 @@ def __init__(self, *args, **kwargs):
...
-                    x = -2
-                self._hash = x
+                if python38:
+                    # From Python 3.8 source code
+                    len_ = len(self)
+                    acc = XXPRIME_5
+                    for i in range(len_):
+                        lane = hash(self[i])
+                        if lane == -1:
+                            return -1
```
klaussfreire: IIUC that should not be necessary, as -1 is an error return code. Cython would already be checking it. Maybe double-check.

sharedbuffers/mapped_struct.py
```diff
@@ -2412,24 +2634,49 @@ def __init__(self, *args, **kwargs):
...
+                        acc += lane * XXPRIME_2
+                        acc = (acc << 31) | (acc >> 33)
+                        acc *= XXPRIME_1
+                    acc += len_ ^ (XXPRIME_5 ^ 3527539)
+                    if acc == -1:
+                        self._hash = 1546275796
+                        return self._hash
```
klaussfreire: Don't return directly, let it fall back to the regular code path:

``` python
if acc == -1:
    acc = 1546275796
self._hash = acc
```

tests/mapped_struct.py
chpedersen: If I do that, `udtype='unicode'` in `mapped_tuple.pack_into` and `mapped_frozenset.pack_into` doesn't work: "Not a type". I couldn't find a workaround.

klaussfreire (changes requested):

tests/mapped_struct.py
klaussfreire: I'm talking only about doing that in tests. There's no cython in tests.

tests/mapped_struct.py
chpedersen: IIUC, if the `_mix_hash` function were something like `code1 ^ code2` it would work, right? (I mean it wouldn't be order dependent, the hash would be different ofc)

```python
def _mix_hash(code1, code2):
    return _hash_rotl(code1, 5) ^ code2
def _stable_hash(key):
...
    elif isinstance(key, (tuple, frozenset, proxied_tuple, proxied_frozenset)):
        if isinstance(key, (frozenset, proxied_frozenset)):
            hval = _FSET_SEED
        else:
            hval = _TUPLE_SEED

        for value in key:
            hval = _mix_hash(hval, _stable_hash(value))
```

klaussfreire: `code1 ^ code2` is just a bad mix function.

Imagine if `l` is some list, any list, `hash(l + l)` is always 0.

I suspect stable_hash is broken for python frozensets... in both versions... huh...

klaussfreire: I'll look into it and maybe make a PR...

sharedbuffers/mapped_struct.py
chpedersen: yep
```c
+2653:                         lane = hash(self[i])
          __pyx_t_6 = __Pyx_GetItemInt(((PyObject *)__pyx_v_self), __pyx_v_i, Py_ssize_t, 1, PyInt_FromSsize_t, 0, 1, 1); if (unlikely(!__pyx_t_6)) __PYX_ERR(0, 2653, __pyx_L1_error)
          __Pyx_GOTREF(__pyx_t_6);
          __pyx_t_7 = PyObject_Hash(__pyx_t_6); if (unlikely(__pyx_t_7 == ((Py_hash_t)-1))) __PYX_ERR(0, 2653, __pyx_L1_error)
          __Pyx_DECREF(__pyx_t_6); __pyx_t_6 = 0;
          __pyx_v_lane = __pyx_t_7;
```

sharedbuffers/mapped_struct.py
```diff
@@ -2614,9 +2607,9 @@ def __repr__(self):
     _cpython = sys.subversion[0] == 'CPython'
 is_cpython = cython.declare(cython.bint, _cpython)
 
-XXPRIME_1 = cython.declare(cython.long, 11400714785074694791)
-XXPRIME_2 = cython.declare(cython.long, 14029467366897019727)
-XXPRIME_5 = cython.declare(cython.long, 2870177450012600261)
+XXPRIME_1 = cython.declare(cython.size_t, 11400714785074694791)
```
klaussfreire: Py_hash_t if it's for python hashes, IIRC.

chpedersen: `Py_uhash_t`

https://github.com/python/cpython/blob/08deed1af56bec8668c6cb4d5cfd89e393e1fe5e/Objects/tupleobject.c#L308

but I can't use it here, I tried `Py_uhash_t`, `cython.Py_uhash_t`, `"Py_uhash_t"` and it doesn't work (unknown type).

klaussfreire: Yes, it seems Cython uses `Py_hash_t` internally in its pxd library, but doesn't export it to be used in user code. Weird.

Sounds like a bug if you ask me.

klaussfreire: Leave it as `Py_ssize_t`, then, which is what `Py_hash_t` actually is after all. It's the same thing, but pyport.h says:

``` C
typedef Py_ssize_t Py_hash_t;
```

and you **can** use Py_ssize_t in cython code

chpedersen: But it's unsigned, `uhash` not `hash`
```c
typedef size_t Py_uhash_t;
```

klaussfreire: The constants are unsigned, but the return type of PyObject_Hash is signed. Whether you use unsigned or signed for intermediates depends on the calculations done, python's tuple code seems to do everything in unsigned and then cast to signed.

`\_(0_0)_/`

klaussfreire: FTR, the original `long` was also signed. Maybe you meant `unsigned long` or, if you want to make sure it's 64 bit, `unsigned long long`.

sharedbuffers/mapped_struct.py
```diff
@@ -2607,9 +2607,9 @@ def __repr__(self):
     _cpython = sys.subversion[0] == 'CPython'
 is_cpython = cython.declare(cython.bint, _cpython)
 
-XXPRIME_1 = cython.declare(cython.size_t, 11400714785074694791)
-XXPRIME_2 = cython.declare(cython.size_t, 14029467366897019727)
-XXPRIME_5 = cython.declare(cython.size_t, 2870177450012600261)
+XXPRIME_1 = cython.declare("unsigned long long", 11400714785074694791)
```
klaussfreire: NB, but FTR there's `cython.ulonglong`

sharedbuffers/mapped_struct.py
```diff
@@ -2412,24 +2626,46 @@ def __init__(self, *args, **kwargs):
...
+                    # From Python 3.8 source code
+                    len_ = len(self)
+                    acc = XXPRIME_5
+                    for i in range(len_):
+                        lane = hash(self[i])
+                        acc += lane * XXPRIME_2
+                        acc = (acc << 31) | (acc >> 33)
+                        acc *= XXPRIME_1
+                    acc += len_ ^ (XXPRIME_5 ^ 3527539)
+                    if acc == -1:
```
klaussfreire: acc is size_t, which is unsigned, but you're comparing to a signed type. To get well defined behavior, you have to cast the literal:

```suggestion
                    if acc == cython.cast(cython.size_t, -1):
```

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
klaussfreire: Ping

sharedbuffers/mapped_struct.py
```diff
@@ -1425,21 +1576,52 @@ def proxied_list_richcmp(a, b, op):
...
         othere = b[i]
 
-        if selfe < othere:
-            return -1
-        elif selfe > othere:
-            return 1
+        if isinstance(selfe, dict):
+            r = dict_cmp(selfe, othere)
+            if r:
+                return r
```
klaussfreire: Why this?

If it's to support comparing lists that contain dicts, py3 doesn't support it, so it's fine if our proxied list doesn't either:

``` pycon
>>> [{3:4}] < [{}]
TypeError: '<' not supported between instances of 'dict' and 'dict'
```

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
chpedersen: Ok... but this breaks `tests.mapped_struct.DictContainerPackingTest`
(also `None` vs `None` works in py2 but not in py3, there's a test that uses this)

chpedersen: Maybe we can skip those tests for py3 (how?)

klaussfreire: > Ok... but this breaks `tests.mapped_struct.DictContainerPackingTest`

Well, the test should not be testing something that's illegal in py3.

>  (also `None` vs `None` works in py2 but not in py3, there's a test that uses this)

Not sure what you mean by that. I can compare None-containing sutff quite well in py3. Do you have a case at hand?

> Maybe we can skip those tests for py3 (how?)

We could. At least those that should not be passing in py3, but lets check that it is actually the case before we start skipping tests.

chpedersen: > Well, the test should not be testing something that's illegal in py3.

See `testPackUnpack` in `DictContainerPackingTest`, it tests that the packed->unpacked `pt`/`pl` are equal to the original values, and they are lists(or tuple) of dicts.

----

In `ProxiedListPackingTest.testProxiedListSpecificEqual`:
```python
        self.assertEquals(self.pack([1, None, 2.0]), (1, None, 2.0))
```

This uses `proxied_list_cmp`
```python
        if selfe < othere:
            return -1
        elif selfe > othere:
            return 1
```

and `selfe` and `othere` are both `None`. In py3 `TypeError: '<' not supported between instances of 'NoneType' and 'NoneType'`

So this is fixed with what I added:
```python
             if selfe is None and othere is None:
                 continue
```

klaussfreire: That case is not about the dict. It's about an element being None directly. It can be implemented easily by just special-casing "selfe/othere is None"

chpedersen: Yep, that's what I did. Sooo, what about the dict stuff?

sharedbuffers/mapped_struct.py
```diff
@@ -1425,21 +1576,52 @@ def proxied_list_richcmp(a, b, op):
...
+        else:
+            if selfe is None and othere is None:
```
```suggestion
            if selfe is othere:
```

It's a bit more general, and if they're the same object they're equal for sure. Might even help performance.

sharedbuffers/mapped_struct.py
klaussfreire: I see the problem.

`proxied_list_richcmp` uses `proxied_list_cmp` to do the comparison, but dicts and frozensets don't have a total order in py3, and it raises an exception rather than invent a total order the way py2 does.

I mantain that we shouldn't be implementing comparisons that aren't valid in py3, so I think the best option is to implement a `proxied_list_eq` specializing in equality comparisons, to avoid depending on the inequalities for tuple values.

Then `proxied_list_richcmp` can use the eq one for `==` and `!=` checks, and `proxied_list_cmp` for the rest.

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
```diff
@@ -1560,21 +1560,23 @@ def proxied_list_richcmp(a, b, op):
         else:
             raise NotImplementedError
 
-    _cmp = proxied_list_cmp(a, b)
     if op == 2: # Py_EQ
-        return _cmp == 0
+        return proxied_list_eq(a, b)
```
chpedersen, line 1564: What's the difference between this and using `proxied_list_cmp` for everything? Is it that we don't use our custom `dict_cmp` when it's not necessary?

klaussfreire: The difference is that we don't need a `dict_cmp` at all and can throw the same exception python dicts throw.

sharedbuffers/mapped_struct.py
```diff
@@ -1432,14 +1621,21 @@ def proxied_list_cmp(a, b):
...
+    for i in range(min(alen, blen)):
         selfe = a[i]
         othere = b[i]
 
-        if selfe < othere:
-            return -1
-        elif selfe > othere:
-            return 1
+        if isinstance(selfe, dict):
+            r = dict_cmp(selfe, othere)
```
klaussfreire: You can now remove dict_cmp and go back to the original implementation. Maybe keep the `selfe is othere` as an optimization.

klaussfreire (changes requested):

klaussfreire (approved):

doc/source/CHANGELOG.rst
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented here.
 
 The format is largely inspired by keepachangelog_.
 
+v1.0.0 - 2022-06-08
```
klaussfreire:
```suggestion
v1.0.0 - Unreleased
```

klaussfreire (approved):

marianoSecure (approved):

sharedbuffers/mapped_struct.py
```diff
@@ -2216,13 +2422,14 @@ def _frozenset_eq(self, x):
...
-        for i in xrange(xlen):
+        for i in range(xlen):
             if self.objlist._c_getitem(i, dcode, objlen, itemsize, offset, _struct, None) not in x:
                 return False
         return True
 
     def __eq__(self, seq):
         if isinstance(self, proxied_frozenset):
+            return self._frozenset_eq(seq)
             return cython.cast(proxied_frozenset, self)._frozenset_eq(seq)
```
a-villagra, line 2433: 🤔

sharedbuffers/mapped_struct.py
```diff
@@ -2352,12 +2560,14 @@ def _subset(self, seq, strict_subset):
...
         if isinstance(self, proxied_frozenset):
+            return self._subset(seq, True)
             return cython.cast(proxied_frozenset, self)._subset(seq, True)
         else:
             return cython.cast(proxied_frozenset, seq)._subset(self, True)
 
     def __le__(self, seq):
         if isinstance(self, proxied_frozenset):
+            return self._subset(seq, False)
             return cython.cast(proxied_frozenset, self)._subset(seq, False)
```
a-villagra, line 2571: yeah 🤔
If this is intended, delete these

PR #95, opened 2022-01-14
Buffer reference handling fixes

klaussfreire:
Fix buffer acquisition error handling to prevent the possibility of
double-release or other buffer reference mismanagement if an error
happens during the GetBuffer call.

### DIFF
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -1590,7 +1590,11 @@ class proxied_list(object):
             if self.pybuf.buf != cython.NULL:
                 PyBuffer_Release(cython.address(self.pybuf))  # lint:ok
                 self.pybuf.buf = cython.NULL
-            PyObject_GetBuffer(self.buf, cython.address(self.pybuf), PyBUF_SIMPLE)  # lint:ok
+            try:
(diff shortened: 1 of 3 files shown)


sharedbuffers/mapped_struct.py
```diff
@@ -1899,10 +1903,11 @@ def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0):
         bitrep_lo=cython.ulonglong, bitrep_hi=cython.ulonglong)
     def unpack_from(cls, buf, offs, idmap = None):
         buf = _likerobuffer(buf)
+        if cython.compiled:
```
gefarion, line 1906: I don't get it. Why moving the PyObject_GetBuffer outside the `try` fix the problem?

klaussfreire: Consider PyObject_GetBuffer raises an error, but leaves the pointer set to non-null.

The error handling code will try to release, but having errored out on GetBuffer means a reference was not counted, so it will be wrong to release it. It may cause a use-after-free bug by releasing the buffer too early.

gefarion (approved):

chpedersen (approved):

PR #96, opened 2022-01-14
CORE-4190: Fix memoryview refhandling

klaussfreire:
Turns out deallocation from `__dealloc__` was having issues. Python 3.4 and above support using a regular finalizer from cython now, `__del__`, so added that one. To keep it functional in older python versions, kept the dealloc around, but it should do nothing if `__del__` ran.

Also some other fixes:
* cast also reshapes, so check ndim
* buffer = memoryview in py3, so `type(x) is buffer` checks work as expected

### DIFF
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -132,19 +132,28 @@ if python3:
     long = int
     basestring = (bytes, str)
 
+_memoryview = cython.declare(object)
+_memoryview = memoryview
+
 @cython.cfunc
 @cython.inline
 def make_memoryview(a):
     if python3:
-        return memoryview(a).cast("B")
+        if type(a) is not _memoryview:
+            a = _memoryview(a)
+        if a.format != 'B' or a.ndim != 1:
+            a = a.cast('B')
+        return a
     else:
         return buffer(a)
 
 if cython.compiled:
     # Compatibility fix for cython >= 0.23, which no longer supports "buffer" as a built-in type
+    buffer = cython.declare(object, buffer)  # lint:ok
     if python2:
-        buffer = cython.declare(object, buffer)  # lint:ok
         from types import BufferType as buffer
+    else:
+        buffer = _memoryview
 
     assert Py_LT == 0
     assert Py_LE == 1
@@ -208,7 +217,9 @@ def _likebuffer(buf):
     """
     Takes a buffer object as parameter and returns a writable object with buffer protocol.
     """
-    if (type(buf) is buffer or type(buf) is bytearray or type(buf) is bytes or
+    if type(buf) is buffer:
+        return make_memoryview(buf)
+    elif (type(buf) is bytearray or type(buf) is bytes or
             isinstance(buf, (ctypes_Array, bytes))):
         return buf
     else:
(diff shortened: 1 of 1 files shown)


gefarion (approved):

sharedbuffers/mapped_struct.py
```diff
@@ -132,19 +132,28 @@
     long = int
     basestring = (bytes, str)
 
+_memoryview = cython.declare(object)
+_memoryview = memoryview
```
chpedersen, line 136: This is to save us the `__Pyx_GetModuleGlobalName` stuff, right?

sharedbuffers/mapped_struct.py
```diff
@@ -132,19 +132,28 @@
...
+            a = _memoryview(a)
+        if a.format != 'B' or a.ndim != 1:
+            a = a.cast('B')
+        return a
     else:
         return buffer(a)
 
 if cython.compiled:
     # Compatibility fix for cython >= 0.23, which no longer supports "buffer" as a built-in type
+    buffer = cython.declare(object, buffer)  # lint:ok
```
chpedersen, line 152: I understand the `cython.declare(object` but why initialize it to "itself"?

sharedbuffers/mapped_struct.py
```diff
@@ -1752,7 +1772,11 @@ def _init(self):
             if self.pybuf.buf != cython.NULL:
```
chpedersen, line 1769: How can this be != null here?

sharedbuffers/mapped_struct.py
```diff
@@ -3476,6 +3505,13 @@ def __dealloc__(self):
                 PyBuffer_Release(cython.address(self.pybuf))  # lint:ok
                 self.pybuf.buf = cython.NULL
 
+    def __del__(self):
+        # For python 3.4+, in earlier versions it will not be called and we'll depend on dealloc
+        if cython.compiled:
+            if self.pybuf.buf != cython.NULL:
+                PyBuffer_Release(cython.address(self.pybuf))  # lint:ok
+                self.pybuf.buf = cython.NULL
+
```
chpedersen, line 3511: I commented out these two `__del__` and it works :shrug:

chpedersen (approved):

sharedbuffers/mapped_struct.py
klaussfreire: Aha

sharedbuffers/mapped_struct.py
klaussfreire: I just want a native global pointer to `buffer`, and I'm calling it `buffer`

sharedbuffers/mapped_struct.py
klaussfreire: If it's being reinitialized (reset)

sharedbuffers/mapped_struct.py
klaussfreire: Sure. you'll have a memory leak though.

sharedbuffers/mapped_struct.py
chpedersen: I mean, you could just do `buffer = cython.declare(object)`, you are setting it right below this

klaussfreire: Both syntaxes do the same, but that one would leave it uninitialized until it's initialized explicitly. The one I used initializes it from the start, right there in the declaration, which allows those variables to be const (not that this one is).

PR #97, opened 2022-03-21
CORE-4190: Fix file leak in object pools

klaussfreire:


tests/pool.py
```diff
@@ -59,6 +59,7 @@ def testPackObject(self):
 
     def testPreload(self):
         p = pool.TemporaryObjectPool()
+        self.addCleanup(p.close, True)
```
chpedersen, line 63: I think you missed one cleanup in `testPack`

chpedersen (changes requested):

sharedbuffers/pool.py
```diff
@@ -327,17 +327,33 @@ def dump(self, fileobj):
...
+            if discard:
+                sections.extend(self.section_freelist)
+                del self.section_freelist[:]
+            else:
+                self.section_freelist.extend(sections)
+                del sections[:]
+
+        for section in sections:
+            fileobj = getattr(section, 'fileobj', None)
+            if fileobj is not None:
```
chpedersen, line 355: When can this be `None`? I don't see it

klaussfreire: Future instances. The plan is to support other types of shared memory that may not be associated to a file object.

chpedersen (approved):

valchulen (approved):

PR #98, opened 2022-03-24
CORE-4190: Py3 perf fixes

klaussfreire:
The many differences between Py3 and Py2 forced a few changes
that derailed performance, at times considerably. These are a few
fixes that recover most of the performance lost:

- cimport array.array to get direct access to the buffer
  for fast index iteration, as regular iteration is quite slow
  in py3
- Precompute bytes([x]) for all x at import time, and use a
  lookup list instead, as constructing the bytes manually, while
  relatively fast, is still slow compared to the py2 idiom it
  replaces and in a critical path
- Use mostly byte typecodes, avoid unicode as it forces an
  encode/decode step that's quite onerous. Still accept them
  from client code, but use bytes internally as much as possible.
- array(dtype, initializer) is very slow in py3. Replace it
  with the equivalent but somehow faster frombytes() when the
  initializer is a memoryview
- Avoid constructing arrays if a memoryview can be used instead,
  as py3 typed memoryviews behave like arrays in any case.
- Use array/memoryview tolist when possible, as it's quite optimized
  and much faster than list(x).
- Build idmapper indexes directly into numpy arrays. Py3 ints are
  slower than py2 ints by virtue of being arbitrary precision. The
  difference is tiny but it adds up. Building the index directly
  into a numpy array with cython's typed memoryviews support for
  direct and native access makes a big difference. Allows building
  big 1M arrays directly instead of using 1k batches, avoiding
  a whole sort step.

### DIFF
--- a/sharedbuffers/mapped_struct.py
+++ b/sharedbuffers/mapped_struct.py
@@ -64,7 +64,7 @@ when explicitly stated.
 from __future__ import division
 
 import struct
-import array
+from array import array
 import mmap
 import numpy
 import tempfile
@@ -122,14 +122,22 @@ npfloat32 = cython.declare(object, numpy.float32)
 npempty = cython.declare(object, numpy.empty)
 npfrombuffer = cython.declare(object, numpy.frombuffer)
 npndarray = cython.declare(object, numpy.ndarray)
+nparray = cython.declare(object, numpy.array)
 
 frexp = cython.declare(object, math.frexp)
 
 ctypes_Array = cython.declare(object, ctypes.Array)
 
+bchr = cython.declare(list)
+
 if python3:
     long = int
     basestring = (bytes, str)
+
+    bchr = [bytes([c]) for c in range(256)]
+else:
+    bchr = list(map(chr, range(256)))
+
 
 _memoryview = cython.declare(object)
 _memoryview = memoryview
@@ -685,8 +693,7 @@ class mapped_tuple(tuple):
         rel_offs = cython.longlong, min_offs = cython.longlong, max_offs = cython.longlong,
         offs = cython.Py_ssize_t, implicit_offs = cython.Py_ssize_t, val_offs = cython.Py_ssize_t,
         i = cython.Py_ssize_t, iminval = cython.longlong, imaxval = cython.longlong, cdtype=cython.uchar, udtype='unicode', l_or_L=cython.bint)
-    def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0,
-            array = array.array):
+    def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0):
         baseoffs = offs
         objlen = len(obj)
         udtype = None
@@ -967,17 +974,16 @@ class mapped_list(list):
 
     @classmethod
     @cython.locals(lrv = list, dchar = cython.char, objlen = cython.Py_ssize_t, absix = object,
-        offs = cython.Py_ssize_t, baseoffs = cython.Py_ssize_t, ix = cython.Py_ssize_t)
-    def unpack_from(cls, buf, offs, idmap = None, array = array.array, klass = list):
+        offs = cython.Py_ssize_t, baseoffs = cython.Py_ssize_t, ix = cython.Py_ssize_t,
+        index = array, i = cython.Py_ssize_t)
+    def unpack_from(cls, buf, offs, idmap = None, klass = list):
         if idmap is not None and offs in idmap:
             return idmap[offs]
 
         baseoffs = offs
         buf = _likerobuffer(buf)
         if python3:
-            dcode = chr(buf[offs])
-            dcode_enc = dcode.encode()
-            dchar = dcode_enc[0]
+            dchar = buf[offs]
         else:
             dcode = buf[offs]
             dchar = cython.cast('const char*', dcode)[0]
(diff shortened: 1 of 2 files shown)


sharedbuffers/mapped_struct.py
```diff
@@ -1001,46 +1007,51 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, klass = list)
                 offs += 8
             w = buf[offs:offs + itemsize * objlen]
             if python3:
-                w = w.cast(dtype)
-            rv = array(dtype, w)
+                rv = w.cast(dtype)
+            else:
+                rv = array(dtype)
+                rv.frombytes(w)
```
chpedersen: Shouldn't it be

```python3
rv = array(dtype)
rv.frombytes(w)
```

for both py2 and py3?

sharedbuffers/mapped_struct.py
```diff
@@ -1001,46 +1007,51 @@ def unpack_from(cls, buf, offs, idmap = None, array = array.array, klass = list)
...
             objlen >>= 8
             offs += 8
             q = buf[offs:offs + itemsize * objlen]
             if python3:
-                q = q.cast(dtype.decode())
-            rv = array(dtype.decode(), q)
+                rv = q.cast(dtype)
+            else:
+                rv = array(dtype)
+                rv.frombytes(q)
```
chpedersen: Same here

chpedersen (changes requested):
Ermm, the tests don't pass for me in py2 :thinking: 

```
======================================================================
ERROR: testIndexedCollectionEqual (tests.mapped_struct.MappedListPackingTest)
----------------------------------------------------------------------
  File "/home/chris/workspace/sharedbuffers/tests/mapped_struct.py", line 1637, in testIndexedCollectionEqual
    c = pack([1, 2.0])
  File "/home/chris/workspace/sharedbuffers/tests/mapped_struct.py", line 1512, in pack
    return self.PACKING_CLASS.unpack_from(a, 0)
  File "sharedbuffers/mapped_struct.py", line 1054, in sharedbuffers.mapped_struct.mapped_list.unpack_from
    index.frombytes(q)
AttributeError: 'array.array' object has no attribute 'frombytes'


======================================================================
ERROR: testIndexedCollectionError (tests.mapped_struct.MappedListPackingTest)
----------------------------------------------------------------------
  File "/home/chris/workspace/sharedbuffers/tests/mapped_struct.py", line 1606, in testIndexedCollectionError
    c = self.pack([])
  File "/home/chris/workspace/sharedbuffers/tests/mapped_struct.py", line 1512, in pack
    return self.PACKING_CLASS.unpack_from(a, 0)
  File "sharedbuffers/mapped_struct.py", line 1013, in sharedbuffers.mapped_struct.mapped_list.unpack_from
    rv.frombytes(w)
AttributeError: 'array.array' object has no attribute 'frombytes'

...
```

klaussfreire:
> Ermm, the tests don't pass for me in py2 thinking
> 
> ```
> ======================================================================
> ERROR: testIndexedCollectionEqual (tests.mapped_struct.MappedListPackingTest)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/chris/workspace/sharedbuffers/tests/mapped_struct.py", line 1637, in testIndexedCollectionEqual
>     c = pack([1, 2.0])
>   File "/home/chris/workspace/sharedbuffers/tests/mapped_struct.py", line 1512, in pack
>     return self.PACKING_CLASS.unpack_from(a, 0)
>   File "sharedbuffers/mapped_struct.py", line 1054, in sharedbuffers.mapped_struct.mapped_list.unpack_from
>     index.frombytes(q)
> AttributeError: 'array.array' object has no attribute 'frombytes'
> 
> 
> ======================================================================
> ERROR: testIndexedCollectionError (tests.mapped_struct.MappedListPackingTest)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/chris/workspace/sharedbuffers/tests/mapped_struct.py", line 1606, in testIndexedCollectionError
>     c = self.pack([])
>   File "/home/chris/workspace/sharedbuffers/tests/mapped_struct.py", line 1512, in pack
>     return self.PACKING_CLASS.unpack_from(a, 0)
>   File "sharedbuffers/mapped_struct.py", line 1013, in sharedbuffers.mapped_struct.mapped_list.unpack_from
>     rv.frombytes(w)
> AttributeError: 'array.array' object has no attribute 'frombytes'
> 
> ...
> ```

Yes, sorry. It's called `fromstring` in py2, I need to do an if python3 there.

sharedbuffers/mapped_struct.py
klaussfreire: No, memoryviews can be iterated as an array, and have a .tolist method, and that's all we care about in this context. And rv.tolist is faster in py3 with memoryviews, so it's preferred as it avoids having to create the intermediate array.

chpedersen (approved):

sharedbuffers/mapped_struct.py
```diff
@@ -6517,48 +6540,43 @@ def build(cls, initializer, destfile = None, tempdir = None,
...
             parts = []
             islice = itertools.islice
-            array = numpy.array
             curpos = basepos + cls._Header.size
-            part = []
             struct_dt = numpy.dtype([
                 ('key', dtype),
                 ('value', dtype),
             ])
+            ppartbuf = partbuf = numpy.empty((1000000, 2), npuint64)
```
gefarion: Why did you choose this buffer size? Can't we use a constant for the buffer size?

gefarion (approved):
Beside the optimization decisions, the code looks good.

sharedbuffers/mapped_struct.py
klaussfreire: It's about the size at which malloc starts using mmap to allocate stuff, which means deallocation actually frees memory immediately to the OS (munmap). Smaller allocations go to the heap and get put on a freelist instead when deallocated, which does not return memory to the OS.

The actual threshold is dynamic but my experience is that it tends to roam the 8MB range.

I'm not sure hypertuning this would help much, the impact isn't huge and the optimal size would depend a lot on the use case.

A constant... well, why not, but I don't have a good name for it. :thinking:

PR #100, opened 2022-06-09
BTS-619: Remove pure python stuff

chpedersen:


sharedbuffers/mapped_struct.py
```diff
@@ -2429,7 +2302,6 @@ def _frozenset_eq(self, x):
 
     def __eq__(self, seq):
         if isinstance(self, proxied_frozenset):
-            return self._frozenset_eq(seq)
             return cython.cast(proxied_frozenset, self)._frozenset_eq(seq)
         else:
```
chpedersen, line 2280: Ok, I don't remember why I added this, but the tests pass :shrug:

sharedbuffers/mapped_struct.py
```diff
@@ -2567,7 +2439,6 @@ def __lt__(self, seq):
 
     def __le__(self, seq):
         if isinstance(self, proxied_frozenset):
-            return self._subset(seq, False)
             return cython.cast(proxied_frozenset, self)._subset(seq, False)
```
chpedersen, line 2416: same here

sharedbuffers/mapped_struct.py
```diff
@@ -1934,46 +1842,35 @@ def __getitem__(self, index):
...
-            return proxied_list_cmp(self, other) < 0
-
-        def __gt__(self, other):
-            if not islist(other):
-                raise NotImplementedError
-            return proxied_list_cmp(self, other) > 0
+    def __cmp__(self, other):
+        if not islist(other):
+            raise NotImplementedError
+        return proxied_list_cmp(self, other)
```
klaussfreire: This was "if not cython.compiled"... shouldn't this have been deleted?

I don't see cython code for the comparison. Which is weird.

I think proxied_list should have a richcmp method like proxied_tuple... not sure why it isn't there.

marianoSecure (approved):

sharedbuffers/mapped_struct.py
```diff
@@ -2675,8 +2546,6 @@ def __hash__(self):
                     if x == -1:
                         x = -2
                     self._hash = x
-            else:
-                self._hash = hash(tuple(iter(self)))
```
klaussfreire: I think the else path is still needed

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
marianoSecure: I think its being hashed correctly in line 2525, since implementation is always going to be cpython

klaussfreire: pypy is not cpython

klaussfreire (approved):

sharedbuffers/mapped_struct.py
klaussfreire: And cython != cpython (one very significant letter of difference)

PR #103, opened 2023-01-20
Store timestamps in UTC

mauguignard:


sharedbuffers/mapped_struct.py
```diff
@@ -2802,7 +2803,7 @@ def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0):
                 widmap.link(objid, obj)
 
         packer = cls.PACKER
-        timestamp = int(time.mktime(obj.timetuple()))
+        timestamp = int(calendar.timegm(obj.timetuple()))
```
klaussfreire: This doesn't work right for datetime, but it does work for date.

I'd say revert datetime, and lets fix date for now.

sharedbuffers/mapped_struct.py
```diff
@@ -2850,7 +2852,7 @@ def unpack_from(cls, buf, offs, idmap = None):
 
         packer = cls.PACKER
         timestamp, = packer.unpack_from(buf, offs)
-        rv =  date.fromtimestamp(timestamp)
+        rv =  date.fromtimestamp(time.mktime(time.gmtime(timestamp)))
```
```suggestion
        rv =  datetime.utcfromtimestamp(timestamp).date()
```

That works reliably.

klaussfreire (changes requested):

sharedbuffers/mapped_struct.py
klaussfreire: Try to add a few tests

sharedbuffers/mapped_struct.py
mauguignard: Done!

sharedbuffers/mapped_struct.py
mauguignard: Done!

sharedbuffers/mapped_struct.py
```diff
@@ -2803,7 +2803,7 @@ def pack_into(cls, obj, buf, offs, idmap = None, implicit_offs = 0):
                 widmap.link(objid, obj)
 
         packer = cls.PACKER
-        timestamp = int(calendar.timegm(obj.timetuple()))
+        timestamp = int(time.mktime(obj.timetuple()))
```
klaussfreire, line 2806: I think it needs to be timegm. There was a case that mktime didn't handle correctly.

mauguignard: This is for the `datetime` (that we talked about leaving it as it was). This commit was just a "rollback". `date` now uses `calendar.timegm`.

klaussfreire: Oh, sorry. Lost track of the context I guess.

klaussfreire (approved):

sharedbuffers/mapped_struct.py
mauguignard: No worries! :raised_hands:

PR #104, opened 2023-01-26
BTS-1230: TZ-aware datetimes

klaussfreire:
Modifies mapped_datetime to make more sense across timezones, and creates new timezone-aware datetime classes.

See changelog entries for more details.


sharedbuffers/mapped_struct.py
```diff
@@ -2789,21 +2799,51 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+        if obj.tzinfo is None:
+            naive_tz = cls.NAIVE_TZ
+            if naive_tz is not None and callable(naive_tz):
+                naive_tz = naive_tz()
+            if naive_tz is not None:
+                obj = obj.replace(tzinfo=naive_tz)
+        if obj.tzinfo is not None:
+            timestamp = int(calendar.timegm(obj.astimezone(TZUTC).timetuple()))
+        else:
+            timestamp = int(time.mktime(obj.timetuple()))
```
mauguignard, line 2844: Shouldn't this use `calendar.timegm` as well? Otherwise the timestamp will be stored in local time

doc/source/CHANGELOG.rst
```diff
@@ -5,13 +5,37 @@ All notable changes to this project will be documented here.
...
+    and stores the original timezone so it can be preserved upon reading
+  - ``mapped_datetime_tz_local`` datetime with timezone that treats naive datetimes as
+    local time.
+
+Changed
+~~~~~~~
+
+- The behavior of ``datetime`` with naive datetime objects should now make more sense,
+  but if an application made use of mixed or non-UTC datetime objects, it may differ from
+  version 1.0.0 in an application-breaking way. It is however recommended that a change to
```
mauguignard, line 31: Wouldn't version `2.0.0` be more suitable due to this?

setup.py
```diff
@@ -129,7 +129,7 @@ def __nonzero__(self):
     extra['ext_modules'] = lazy_modules()
     extra['setup_requires'] = setup_requires
 
-VERSION = "1.0.1"
+VERSION = "1.1.0"
```
mauguignard, line 132: There is a `_version.py` file as well:

https://github.com/jampp/sharedbuffers/blob/0ee10a437cb8500f13817e258c96a30c8bcdc165/sharedbuffers/_version.py#L1

Maybe we should unify it :thinking:

klaussfreire: _version is generated from VERSION when you run the build, I haven't run it since bumping the version file it seems.

doc/source/CHANGELOG.rst
klaussfreire: We made no guarantees in 1.0.0 and the changes aim to fix unintuitive/broken behavior rather than a proper contract change, while keeping common usage the same. I would frame this more as a fix than an API change, which is why I didn't go for 2.0.0

sharedbuffers/mapped_struct.py
klaussfreire: [Unix time](https://en.wikipedia.org/wiki/Unix_time) is specified as UTC, so mktime should be converting to GMT assuming the timetuple is in localtime. In fact it does, and all our problems stem from that little hidden detail.

In contrast, FTR, `timegm` assumes the timetuple is in GMT.

sharedbuffers/mapped_struct.py
```diff
@@ -2817,12 +2857,117 @@ def unpack_from(cls, buf, offs, idmap = None):
...
+                tz_info = TZUTC
+            tz_info = tz_cache.setdefault(utc_off, tz_info)
+        rv = datetime.fromtimestamp(timestamp >> 20, tz_info) + timedelta(microseconds=microseconds)
+
+        if idmap is not None:
+            idmap[offs] = rv
+        return rv
+
+
+class mapped_datetime_tz_local(datetime):
```
gefarion: Is it the parent class ok? Don't have to be mapped_datetime_tz?

klaussfreire: No, it's a complete implementation so I made it separate.

klaussfreire: Wait, this is the tz_local, yes

gefarion (approved):

mauguignard (approved):

